Issue #491, opened 2014-05-19
Multiple assignment doesn't work with index operator


ephere:
Test case:

```
using System;

public class T
{
    public float[] values = new float[3];

    public float this[int x]
    {
        get { return this.values[x]; }
        set { this.values[x] = value; }
    }
}

public static class Program
{
    public static void Main( string[] args )
    {
        var t1 = new T();
        t1[0] = t1[1] = 1.0f;

        Console.WriteLine( t1[0] );
    }
}
```

Result:

```
Result Message: 
Expected string length 1 but was 3. Strings differ at index 0.
  Expected: "1"
  But was:  "NaN"
  -----------^
```

The JS output responsible for the assignment is:

```
function Program_Main (args) {
    var t = new ($T01())();
    t.set_Item(0, t.set_Item(1, 1));
    $T02().WriteLine(t.get_Item(0));
  };
```

iskiselev:
Unfortunately it is still not full fix. Here is new problem test case:
```
using System;

public static class Program {
    public static void Main (string[] args)
    {
        int i = 1;
        Test(ref i);
    }

    public static void Test(ref int i)
    {
        var array = new[] { 1, 2, 3 };
        Console.WriteLine(array[i++]);
    }
}
```
And after translation:
```
  function Program_Test (/* ref */ i) {
    var $temp00;
    var array = JSIL.Array.New($T01(), [1, 2, 3]);
    $T03().WriteLine(array[($temp00 = (i.get() | 0), 
        i.get() = (((i.get() | 0) + 1) | 0), 
        $temp00)]);
  };
```

In JS strict mode (#677) such problem (assign to non L-value) throw exception on script load.

iskiselev:
Opened #684 for issue in previous comment, as it is anther problem.

PR #614, opened 2014-10-17
Fix assignment of indexer setters

iskiselev:
Fix assignment of indexer setters (#491) and test added.
Small refactoring for char arithmetic decompress. Now we use DecomposeMutationOperators for it.


iskiselev:
Close this PR as it was resolved in right way by @kg (see #622 for comments).

Issue #619, opened 2014-10-20
Internalizer with read only property 


iskiselev:
Internalizer with read only property doesn't work - it fails with "Property is not writable". Here is test case:

```
using System;

public class ClassWithReadOnlyProperty
{
    private ComplexProperty _property = new ComplexProperty {SomeString = "Initial"};

    public ComplexProperty Property
    {
        get { return _property; }
    }
}


public class ComplexProperty 
{
    public string SomeString { get; set; }
}

public static class Program {
    public static void Main (string[] args) {
        var instance = new ClassWithReadOnlyProperty
        {
            Property = { SomeString = "Modified" }
        };

        Console.WriteLine(instance.Property.SomeString);
    }
}
```

I'll look on this issue later today.

iskiselev:
Works, test to be added?

PR #620, opened 2014-10-20
Fixed initializer for read only properties

iskiselev:
Fix for (#619).
Not best implementation - really it would be better don't write `JSIL.ObjectInitializer` at all if we create new object and use it only if we just modify properties.


iskiselev:
Closed as @kg implemented right solution (see comments in #622).

PR #622, opened 2014-10-20
Expression Interpreter updated with #491 fix.

iskiselev:
Updated version of Expression Interpreter with #491 fix.

### DIFF
--- a/Libraries/JSIL.ExpressionInterpreter.js
+++ b/Libraries/JSIL.ExpressionInterpreter.js
@@ -38283,7 +38283,10 @@ JSIL.MakeEnum(
   function LightCompiler_EnsureLabel (node) {
     var result = new JSIL.BoxedVariable(null);
     if (!$T05().prototype.TryGetValue.call(this._treeLabels, node, /* ref */ result)) {
-      result.set($T05().prototype.set_Item.call(this._treeLabels, node, new ($T31())(node)));
+      var _ic_0 = null;
+      result.set((_ic_0 = new ($T31())(node), 
+          $T05().prototype.set_Item.call(this._treeLabels, node, _ic_0), 
+          _ic_0));
     }
     return result.get();
   };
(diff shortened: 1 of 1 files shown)


kg:
Sorry to do this, but could you possibly redo this and use the current emscripten branch? I removed object/collection initializers (it makes a bunch of codegen improvements possible) and I can't kill __Initialize__ until expression interpreter is updated anyway :-)

If it's too much trouble I can probably do it myself (let me know how you generate this file, perhaps?)

Thanks

iskiselev:
I'll prepare new pull request in few days.
Probably it will be dependency to my repo with ExpressionIterpreter prepared as .dll project, that is able to be translated only if needed.
Let's keep this PR open as a reminder for me.

kg:
Thanks. No rush.

Happy to help introduce automation for this into the JSIL build pipeline, if it helps.

iskiselev:
@kg, for better understanding - what are plans for emscripten branch? Will it be merged later with master?

kg:
Yes, I'm probably merging it soon.

iskiselev:
I've checked - currently #491 is still blocking JSIL from correctly translation Expression Iterpreter to JS. I've prepared PR #614 some time ago, that used not best, but mostly workable solution for it.

We should solve it first - but now I'm only able to prepare translation of it myself with patched JSIL version.

So, if you want - I'll do new JSIL.ExpressionInterpreter.js, but for translation it only if requires we need solve #491 first.

kg:
Thanks, I'll try and fix 491 on the emscripten branch. Hadn't realized it was blocking you.

kg:
02d0312811b4341479bfc5baf4631c534abf4967 should do the trick. Let me know.

iskiselev:
Looks like suggested fix doesn't cover all cases (I've not checked if it covers enough cases to unblock me).
Can you look on test case that I've write for PR #614. Here is direct link:
https://github.com/iskiselev/JSIL/blob/IndexerAssigments_Issue491/Tests/SimpleTestCases/MultipleIndexerAssgments_Issue491.cs

kg:
```javascript
    variable = +($temp02 = +t.get_Item(0) + 50, 
      (t = store).set_Item(0, $temp02), 
      $temp02);
```

... I hate ILSpy so much. Thanks, interesting test case.

kg:
Boy that was complicated.
Try 8eb81cac35ffb7bdd00bb1f438b5a9e1705c1e82

iskiselev:
Unfortunately it is still generate broken code if we set type that is generic.
Here is an example:
```
using System.Collections.Generic;

public static class Program
{
    public static Dictionary<string, Generic<int>> _runCache;

    public static void Main()
    {
        Run();
    }

    public static Generic<int> Run()
    {
        return _runCache[string.Empty] = new Generic<int>();
    }
}

public class Generic<T>
{
}
```

It is translated with such script:
```
  function Program_Run () {
    var $temp00, $temp01, $temp02;
    return ($temp00 = $thisType._runCache, 
      $temp01 = $T00().Empty, 
      $temp02 = new ($T01())(), 
      $temp00.set_Item($temp01, JSIL.CloneParameter(TValue, $temp02)), 
      $temp02);
  };
```

There is no `TValue`, that is passed to JSIL.CloneParameter.

kg:
Fixed by 2d15640155816054898ff431cb0f908994a5b85d

iskiselev:
Close this PR, as we should work on updated version in #683.

PR #656, opened 2015-01-31
Phonegap application

jwaynelee:


iskiselev:
@kg, what do you think about this PR?
Support of web worker is really good JSIL, other changes also looks good.
I could prepare new version for it on top of current head.

kg:
I was confused by the title so I never looked at it. The contents seem fine to me

PR #685, opened 2015-02-13
Ref Integer increment issue fixed.

iskiselev:
Fix for #684 and test added.

Issue #677, opened 2015-02-12
Always use JS strict mode


iskiselev:
As per @kg comment in #624, we should always use JS strict mode. Currently JSIL generated code is not fully compatible with strict mode. Here is list of already found issues:
1. String that contains null byte following digital literal. Really it was JSIL bug - so `"\01" was translated  to `"\01"` string, which means sting that contains 1 symbol and this symbol - is byte 1, instead of string that contains null byte following by character 1. In strict mode  such sequence is illegal
2. We should rename function parameters or variables inside function with name `arguments`.

Some other error on which I'm still working. (@kg, I'll prepare PR for all of it).

kg:
Good points on both. I think renaming 'arguments' is pretty easy, there's already a reserved words list somewhere that contains most of the other JS ones. Using \x00 always seems like a reasonable solution to that problem.

iskiselev:
One more issue:
When CodeGenerator.EnableUnsafeCode = true (JSIL default setting), next code:
```
using System;

public static class Program
{
    public static void Main(string[] args)
    {
    }

    private static unsafe void MDTransform(uint* state)
    {
        uint num2 = state[1];
    }
}
```
Is translated with:
```
  function Program_MDTransform (state) {
    var num2 = ((state + 4).get() >>> 0);
  };
```

Which is incompatible with strict mode. It is problem, as BCL has such code and BCL translation with  default settings is unable to load.

kg:
Oh weird, that one's definitely a bug. It should be calling .add() on state, or getElement/getOffset.

iskiselev:
It call getOffset, but only when EnableUnsafeCode = true, otherwise it generates bad code.

kg:
Oh you misstated that originally, you said when EnableUnsafeCode = true, but you mean that bad code is generated when it's set to false? That should be fixable then, thanks

iskiselev:
One more issue:
```
using System;

public static class Program
{
    public static void Main(string[] args)
    {
    }

    private static unsafe void SHATransform(ulong* state)
    {
        var b = state[1];
        state[1] += b;
    }
}
```
Produce:
```
  function Program_SHATransform (state) {
    var b = state.getOffset(8);
    state.getOffset(8) = $T01().op_Addition(state.getOffset(8), b);
  };
```

Note: There will be no problem if we change type from UInt64 to UInt32 - probable `DecomposeMutationOperators.ConvertReadExpressionToWriteExpression` do this trick then.

[...]

PR #690, opened 2015-02-17
Strict mode

iskiselev:
Activated JS Strict mode for all generated files (#677).
This PR should be applied after #685, #688, #689 - otherwise some tests will fail.

### DIFF
--- a/JSIL/AssemblyTranslator.cs
+++ b/JSIL/AssemblyTranslator.cs
@@ -572,6 +572,7 @@ namespace JSIL {
             using (var ms = new MemoryStream())
             using (var tw = new StreamWriter(ms, new UTF8Encoding(false))) {
                 tw.WriteLine("// {0} {1}", GetHeaderText(), Environment.NewLine);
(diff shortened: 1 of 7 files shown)


iskiselev:
Applied fixes:
1. We should rename function parameters or variables inside function with name arguments
Solution: `arguments` was added to Util.ReservedWords
2. `var num2 = ((state + 4).get() >>> 0);` when EnableUnsafeCode = false
Solution: EnableUnsafeCode = true for BCL translation. Probably we should implement proper workaround.
3. JSIL.IgnoredMember("Field", $thisType) = 1;
Solution: generate code `JSIL.IgnoredMember("Field", $thisType).LValue = 1;`. As JSIL.IgnoredMember will abort script execution in any case, we should not have any additional error that we have no LValue in this function result.

kg:
Those all seem good to me. I'll have to merge this manually

iskiselev:
I've rebased PR, so there is no merge anymore.

PR #698, opened 2015-03-04
Delegate update

iskiselev:
Two updates for delegate:
1. Support unbind form of delegate to instance method (so that we can pass instance as first argument).
2. Disabled guess of proper MethodInfo for interface delegate, as it works only in some conditions. We need to find beter solution here (we need some way to find method info in implementation based on methodinfo in declaration and implementation type). This logic should be written similar to the logic, that select correct method on call - but we need discuss how to reuse already existing code or refactor it for reuse.

This change need for interpreting some more complex expressions.


kg:
Very comprehensive tests, great stuff

PR #699, opened 2015-03-04
Updated JSIL.ExpressionInterpreter with proper .gitignore

iskiselev:
Fix iskiselev/JSIL.ExpressionInterpreter#1.


kg:
I think I already did this.

PR #703, opened 2015-03-04
Fixed error on reflection Invoke of static method from generic class

iskiselev:
Fixed error on reflection Invoke of static method from generic class and test added (I've not created issue for it).
Added op_Equality and op_Inequality for ConstructorInfo.

Sorry, I deleted branch from old #616 PR while tried to rebase it. It is totally same - but I was unable to reopen old PR.

@kg, you haven't merged it for a while - do you want to change something here?


kg:
I haven't the faintest idea why I never merged this. I'll take another look.

PR #704, opened 2015-03-05
ReadOnlyCollection creation with IList<T>. Test added.

iskiselev:
Previous implementation of ReadOnlyCollection constructor suppose that input is array or List<T> - so custom IList<T> implementation doesn't work.
It is fixed here (I've not created separate issue for it). This fix is last necessary fix for Expresssion Interpreter correct work.


kg:
Should creating a ReadOnlyCollection from an IList make an array copy like that? I think it should delegate to the IList instead.

iskiselev:
You are right. This implementation is incorrect. You may close this PR - I
will do a new one when do it right.

PR #707, opened 2015-03-10
Failed to extract BoxedVariable before return result

### LINKED ISSUE #705: Failed to extract BoxedVariable before return result
JSIL in some cases doesn't extract value from BoxedVariable before return it as non-ref function result. Here is test case:
```
using System;

public static class Program {
    public static void Main(string[] args) {
        Resolver.GetResult().RunMe();
    }
}

public static class Resolver
{
    private static object _syncRoot = new object();

    internal static ResultClass GetResult()
    {
        lock (_syncRoot)
        {
            ResultClass factory;
            if (!TryGet(out factory))
            {
                factory = Create();
            }

            return factory;
        }
    }

    private static ResultClass Create()
    {
        return new ResultClass();
    }

    private static bool TryGet(out ResultClass result)
    {
        result = null;
        return false;
    }
}

public class ResultClass
{
    public void RunMe()
    {
        Console.WriteLine("Hura!");
    }
}
```

Problem is inside `GetResult` method. Here is its translation:
```
    var factory = new JSIL.BoxedVariable(null);
    var syncRoot = null;
    var flag = new JSIL.BoxedVariable(false);
    try {
      $T03().Enter(syncRoot = $thisType._syncRoot, /* ref */ flag);
      if (!$thisType.TryGet(/* ref */ factory)) {
        factory.set($thisType.Create());
      }
    } finally {
      if (flag.get()) {
        $T03().Exit(syncRoot);
      }
    }
    return factory;
```

There will be no problem, if we remove lock from `GetResult` method.

iskiselev:
Fix #705. Root cause of problem:
C# lock statement use temporary variable inside.
During transformation pass EliminateTemporaries, this temporary variable removed inside `EliminateSingleUseTemporaries.EliminateVariable` with next code:
```
var replacer = new VariableEliminator(
                    variable,
                    JSChangeTypeExpression.New(replaceWith, variable.GetActualType(TypeSystem), TypeSystem)
                );
```
Than, nobody replace JSChangeTypeExpression with any other expression and `JavascriptAstEmitter` simple write inner expression, ignoring type changing:
```
public void VisitNode (JSChangeTypeExpression cte) {
            Visit(cte.Expression);
        }
```

I'm not sure, that I've fixed problem in best place - just added creation of `JSReadThroughReferenceExpression` inside `JSChangeTypeExpression.New` function. Maybe it should be fixed in another way.

### DIFF
--- a/JSIL/AST/JSExpressionTypes.cs
+++ b/JSIL/AST/JSExpressionTypes.cs
@@ -2305,6 +2305,7 @@ namespace JSIL.Ast {
         public static JSExpression New (JSExpression inner, TypeReference newType, TypeSystem typeSystem) {
             var cte = inner as JSChangeTypeExpression;
             var literal = inner as JSIntegerLiteral;
+            var variable = inner as JSVariable;
             JSChangeTypeExpression result;
 
             if (cte != null) {
@@ -2318,9 +2319,14 @@ namespace JSIL.Ast {
             var innerType = inner.GetActualType(typeSystem);
             if (TypeUtil.TypesAreEqual(newType, innerType))
                 return inner;
+            else if ((variable != null) && innerType.IsByReference &&
+                     TypeUtil.TypesAreEqual(((ByReferenceType) innerType).ElementType, newType)) {
+                return new JSReadThroughReferenceExpression(variable);
+            }
             else if ((literal != null) && TypeUtil.IsPointer(newType)) {
                 return new JSPointerLiteral(literal.Value, newType);
-            } else
+            }
+            else
                 return result;
         }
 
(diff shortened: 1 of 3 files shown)


kg:
I think the fix is probably to never eliminate ref temporaries... but I thought I did that already. I'll have to look into this, thanks.

kg:
Oh I see, the cast expression has to be a read-through-reference, that's probably right. Good catch.

kg:
I've been thinking that JSIL's AST probably needs to explicitly represent lvalues and rvalues as categories, now that there are a few types (references, pointers, etc.)
This is a good reason to consider that more.

PR #709, opened 2015-03-10
string.Compare against null.

iskiselev:
Fixed behavior for string.Compare(a, b) if a or b is null. (I haven't created issue for it).


kg:
Thanks, what a stupid oversight on my part :-)

Issue #713, opened 2015-03-11
MultiThread race condition problems


iskiselev:
I've noticed, that JSIL works very unstable if it runs with multithread support.
I will post descriptions of found problems (I compare bad and good translations) in this issue.

iskiselev:
First issue:
```
var $asm12 = JSIL.GetAssembly("System, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089");
var $asm12 = JSIL.GetAssembly("System.Xml, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089");
```

JSIL used same identifier for two different assemblies.

kg:
I've been using everything with threads enabled by default since I added them, and haven't hit anything, but I don't doubt that it's buggy as hell :-)

The deadlock tracking locks are specifically to resolve a threading issue, so happy to address any other related problems.

iskiselev:
Currently, most of builds of my project translated with threads enabled are corrupted, but I've never seen dead locks.

kg:
How long does a build take with threads vs nothreads? How many assemblies? How big's the output?

Maybe the goal with threads should be for the output to always be identical to nothreads, and then I can just hammer at running testcase compiles in a loop until they fail...

iskiselev:
I have:
-18 translated assemblies from my project 
-JSIL.ExpressionInterpreter
-System.Core
-stubs of mscorlib, System, System.Runtime.Serialization, System.Xml (total 4)

DCE is enabled and took about 7 seconds.

Total size of 18 assemblies - 1.53Mb. Their translated size - 4.65Mb. As DCE is enabled, we really can't compare translated and untranslated size.
Total JSIL translation assemblies output size 9.87Mb plus 1.37Mb of translated string resources.

Total JSIL time in single thread took 28-29 seconds and with UseThreads it took only 16-17 seconds.

iskiselev:
Next issue was already described in #471. Sometimes JSIL skip declaring variable in threads mode. With enabling JS strict mode (#677)  all such problems are critical now.

Here is example from XmlSerializer library translation:
Ilspy decompile
```
private SearchResultWrapper Read38_SearchResultWrapper(bool isNullable, bool checkType)
{
	XmlQualifiedName xmlQualifiedName = checkType ? base.GetXsiType() : null;
	bool flag = false;
	if (isNullable)
	{
		flag = base.ReadNull();
	}
[...]

PR #714, opened 2015-03-11
AssemblyId generation use lock.

iskiselev:
Fix first issue from #713.


kg:
Ahhhhh. Yeah, this is busted. I'm not sure I'll use this fix exactly.

PR #720, opened 2015-03-25
Emscripten test for reading string as ref argument

mispy:
used in FNA for calling glShaderSource
https://www.opengl.org/sdk/docs/man/html/glShaderSource.xhtml


mispy:
not much luck with this one; I get garbled strings which look like uninitialized memory. Tried having RefMarshaller check for a string and wrap it in an extra pointer but similar result

PR #733, opened 2015-04-21
Fix declaring reference to built-in types from enclosed type with DCE

### LINKED ISSUE #727: DCE creates .JS files that won

iskiselev:
Fix #727.
JavascriptFormatter tried to use `$.TypeName` syntax for standard types in enclosed type context. If parent type was stripped to DCE, context is not enclosed anymore - so `$` is not defined.
@kg, probably it will be better always  pass $ as argument to `CreateEnum`/`CreateDelegate` and remove check for enclosed type? I had not tested it much, so don't know yet if there will be other problems.

### DIFF
--- a/JSIL/AssemblyTranslator.cs
+++ b/JSIL/AssemblyTranslator.cs
@@ -884,7 +884,7 @@ namespace JSIL {
 
             var tw = new StreamWriter(outputStream, Encoding.ASCII);
             var formatter = new JavascriptFormatter(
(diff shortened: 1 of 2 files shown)


kg:
Hm, I think I agree with you. I'll think about it, thanks.

iskiselev:
@kg, what about this fix? I've not tested only idea with removing check for enclosed type, but this fix should work fine.

kg:
It looks complicated so I don't want to merge it until I understand it. Sorry.

PR #739, opened 2015-06-14
Use Protobuild for project definitions and package management

hach-que:
[Protobuild](http://protobuild.org/) is a project generator for cross-platform projects.  It is used by MonoGame and other projects to support cross-platform code.

This pull request replaces the C# solution and project files currently residing in the JSIL repository with a Protobuild module.  JSIL can then be targeted to various desktop platforms with `Protobuild.exe --generate Linux` (or by double-clicking the executable on Windows).

Why do this?

**Initial cross-platform support for JSIL**

Prior to this you could not build JSIL out of the box on Linux, because of various projects containing Windows-specific build steps.  Because Protobuild is a project generator targeting a specific platform, projects no longer contain platform-specific build steps.

This also means that the XNA projects are only generated and included in the solution when targeting Windows; there is no longer any need for the JSIL_NoXNA.sln file.

There's still more we could do here though, with regards to changing some of the code when JSIL is built for Mac OS / Linux (there are some things such as concurrency which Mono doesn't handle very well).

**Submodule caching**

By specifying the Git submodules as Protobuild packages (as shown in https://github.com/hach-que/JSIL/blob/master/Build/Module.xml), they are locally cached on the developer's computer when Protobuild is run.  This means that if you clone JSIL multiple times, which is common on build servers, the submodules are not cloned for each build.  Instead the submodule will be cloned from a local cache (which has `git fetch` called on it first).

**Simplification for the Web platform in Protobuild itself**

One of the platforms Protobuild targets is the Web platform, and it uses JSIL to do this.  Previously Protobuild downloaded and compiled JSIL locally on the developer's computer when they first targeted the Web platform, but this was always kind of kludgy and had various workarounds (such as the need to patch project files so things would build on Linux).  Since that was originally implemented, Protobuild now supports packages, which is a far better way of including JSIL as a tool for use during builds.

This makes using JSIL in Protobuild extremely fast and easy.  This is now what happens when you target the Web platform *without* having JSIL installed on a machine when you start:

output

By merging this PR into upstream, it means that my build server can just point at the upstream repository to build and publish the JSIL compiler package, which means the latest version will always be available (in binary form) to everyone.


iskiselev:
Is I correct, that main benefit for JSIL itself for now from using Protobuild is support of Xamarin Studio as development IDE? What other benefits will JSIL have from this PR?

hach-que:
That, the submodule caching and that it also makes it extremely easy for people to start using JSIL.

PR #741, opened 2015-06-20
Failing test for union out parameters

mispy:
likely related to SDL_Event issues


mispy:
// Generated javascript ends here //
NUnit.Framework.AssertionException:   String lengths are both 12. Strings differ at index 6.
  Expected: "i=6 f=3.7000"
  But was:  "i=6 f=0.0000"
  -----------------^

PR #744, opened 2015-07-03
Test + fix for out ints larger than a byte

mispy:


kg:
Thanks. Pointer code is so goofy.

Issue #240, opened 2013-08-06
No GetHashCode() in object


iskiselev:
There is no any implementation of GetHashCode in JSIL object. 
Would it be correct to add:

        [JSIsPure]
        [JSReplacement("JSIL.ObjectHashCode($this)")]
        [JSRuntimeDispatch]
        public new abstract int GetHashCode();

to JSIL.Proxies.ObjectProxy?

Issue #446, opened 2014-04-26
Trouble with simple boxing


igor-kostromin:
In http://jsil.org/try :

``` csharp
using System;

public static class Program {
  public static void Main () {
    int a = 4;
    int b = 4;
    if (!((object)a).Equals(b))
      Console.WriteLine("Not equals");
    else
      Console.WriteLine("Equals");
  }
}
```

Expected: Equals
But was:

> Unhandled exception: TypeError: undefined is not a function
> TypeError: undefined is not a function
> at Program_Main (eval at <anonymous> (http://jsil.org/try/output_frame.js:70:10), <anonymous>:23:14)
> at runScript (http://jsil.org/try/output_frame.js:80:5)
> at runTranslatedJS (http://jsil.org/try/index.js?v6:597:20)

Can I use any workaround for this ? Actually I need to implement simple method:

``` csharp
public static void AreEqual( object a, object b ) {
    if (!a.Equals( b ))
        throw new InvalidOperationException("Assertion failed.");
}
```

Issue #241, opened 2013-08-06
No Equals for simple type


iskiselev:
In next example:

            int i1 = 1;
            int i2 = 2;
            bool r = i1.Equals(i2);
            Console.WriteLine(r);

JSIL generates next source: 

    var i2 = 2;
    var r = (1).Equals(i2);
    $T02().WriteLine(r);

But 2 is simple type without Equals method. So there is exception: 
Uncaught TypeError: Object 1 has no method 'Equals'

PR #749, opened 2015-07-16
Object virtual methods

iskiselev:
Looks like I've found good enough solution for calling virtual object methods (Equals, GetHashCode) on raw JS types (string/int/double and so on).
For it I've added additional pseudo-calls to `JSReplacement` meta attribute:
- `$explicitThis(method)` - so we can pass information if we should do virtual call
- `$declaringType(method)` - so we can know method declaration type for non-virtual call.

With help of this new parameters, I changed instance virtual object methods with calling JSIL special helper functions, that know how to process raw-js object scenarios.

Test added.

After applying it we can remove hack with adding GetHashCode and Object_Equals method to string prototype - but for simplifying this PR I've not done it here.


iskiselev:
This should resolve #240 , #241, #446 .

kg:
Sounds awesome, I'll take a look later.

Issue #731, opened 2015-04-19
DCE prunes pinvoke methods and related types in some scenarios


kg:
For some reason DCE doesn't seem to be visiting pinvoke methods. As a result this means that custom pinvoke marshaler types get killed by the analyzer. I spent an hour or so trying to fix this and couldn't figure out how to fix it - the DCE code is nigh-incomprehensible, I can't figure out where and why it's pruning types. It seems like it skips methods and types in a dozen different places based on different criteria.

iskiselev:
Can you provide test example?
DCE walks through all types/methods from starting point, skipping types/methods marked with Ignored/External meta attributes.
After this pass, it will says to JSIL skip translations of all types/methods that are not in list from previous step.
Central point here is `DeadCodeInfoProvider.WalkMethod(...)` and`DeadCodeInfoProvider.AddMethod(...)`. First of it is responsible for collecting all references from method body IL stream, second - for adding all dependencies from method declaration (instance type, argument types, return type, generic arguments).

kg:
I tried using AddType to add the type referred to by the PInvoke method, but that didn't seem to affect whether it was walked or not. I tried various things and nothing worked, with the exception of adding a blanket whitelist entry (```Namespace\.TypeName::.*```).

iskiselev:
I've already thought about implementing some log for DCE - in that case it will be easier to understand why some type/method was included (what other type reference it) or why it was excluded (really it should be only if method/type was marked as `IsIgnored` or `IsExternal` or `IsStubOnly`).

And DCE always ignores every method that is not reachable from Starting Point/white listed points.

Currently DCE could not look through custom method implementation in Proxies/FakeMethods - so probably you have such case. DCE is working only on origanal assembly and read  `IsIgnored, `IsExternal`, `IsStubOnly` properties.

kg:
Specifically the problem was with MarshalAs and custom marshalers:
https://msdn.microsoft.com/en-us/library/s2t3h416%28v=vs.90%29.aspx

The attribute referred to a marshaler type, but no matter what I did I couldn't exempt that marshaler from DCE. Adding a whitelist regex would fix it but the regex made compilation like 50x slower.

iskiselev:
Adding whitelist regexp should affect only on DCE analyze step. Is it do so big performance hit?
How long is your full translations takes?

On my project DCE analyze takes about 7 s seconds and saves me several minutes of compilation (I use lot's of regex white list records).

[...]

PR #759, opened 2015-07-21
Dce improvements

iskiselev:
Added test framework for future DCE tests, added DCE basic functions test.
DCE is now capable to eliminate Property/Event definition neither methods of which are used.


kg:
Thanks for adding tests for this. I'll review the PR later.

iskiselev:
Really I'm still working on test coverage and DCE fixes/improvements.

iskiselev:
I've done with current iteration of improving DCE. Most features are covered with tests (may be introduce some more later). Fixed lots of small bugs.
Totally rewritten work with CustomAttributes - now should work good (#731 should be fixed, but may not). DCE eliminates all attributes that are not used in code (writing typeof(AttributeClass) somewhere should be enough).
Added (and make default) AggressiveVirtualFunctionsRemove mode - see PreserveVirtualMethodFromReallyUsedRootOnly test case.

I've found a problem in DCE (see #763), that I've not fixed yet - before we discuss root cause.

Issue #763, opened 2015-07-23
New AssemblyResolver / CachingMetadataResolver on each assembly


iskiselev:
While debugging DCE, I've noticed that we can't compare TypeDefinition for equality (two TypeDefinition that represent same type may have different instance, that will be not equal as it do not override Equals or ==).
Looks like I know root cause of this:
JSIL creates new pair of AssemblyResolver / CachingMetadataResolver for each assembly resolving.
As CachingMetadataResolver has private Metadata cache, we ends with new TypeDefinition for each resolution of TypeReference from different assembly (even if target type is same).

When I've done a dirty hack to always use same AssemblyResolver / CachingMetadataResolver my problem have gone away.

Was it done on purpose or it is just a small bug that should be fixed?

PR #765, opened 2015-07-25
Same caching metadata resolver

iskiselev:
@kg, please answer #763. This is suggest patch if it was a bug.

### DIFF
--- a/Compiler/Program.cs
+++ b/Compiler/Program.cs
@@ -587,7 +587,7 @@ namespace JSIL.Compiler {
             }
 
             var translator = new AssemblyTranslator(
-                configuration, typeInfoProvider, manifest, assemblyCache, 
+                configuration, typeInfoProvider, manifest, new AssemblyDataResolver(configuration, assemblyCache), 
                 onProxyAssemblyLoaded: (name, classification) => 
                     Console.Error.WriteLine("// Loaded proxies from '{0}'", ShortenPath(name))                
             );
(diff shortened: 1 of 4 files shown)


kg:
That sounds like a bug. The intent of the caching resolver is to only do it once ever, I think?

There might be a reason you can't, like per-assembly things that influence assembly resolution? A CLR2 assembly pulls in mscorlib 2.0 and a CLR4 assembly pulls in 4.0? But I'm not sure we care, since mixing corlibs is a huge problem as-is and I don't even know how it works on native .NET :-)

kg:
I notice you touched the ComparisonTest initialization. Are you sure you meant to do that? The cache reuse from before was a huge optimization for tests and it looks like you're changing it to create a new instance of something for every test run instead of re-using an existing one.

iskiselev:
@kg, in this PR I've tried to fully preserve logic of test - earlier tests shared `AssemblyCache`, but each of it used each own CachingMetadataResolver/AssemblyResolver, as it was created inside `GetReaderParameters.AssemblyTranslator` (once per assembly load). Now it should work totally same - new instance of this classes are created once per test, but still share `AssemblyCache` between test.

But while I've typed it, I found one potential problem - we still may use different `CachingMetadataResolver` in testing environment - if mscorlib was loaded in earlier test, but probably it is much less critical - as JSIL code use special logic for comparing `MemberDefinition`, and DCE tests are isolated, but according to this it still may be a good idea implementing custom comparer for them.

iskiselev:
Other option may be changing test logic and also use same `CachingMetadataResolver` in tests. The only problem of this decision that I see is increase of memory usage by tests. This could be resolved by implementing Assembly unload option.

PR #767, opened 2015-07-26
Dce improvements (continue)

iskiselev:
Fixed processing of method arguments, generic parameters, return value and class generic parameters custom attributes (earlier was processed only if method/class has custom attributes).

Added initial support for MarshalInfo (#731) - for now added only CustomMarshalInfo (other may be easy added if needed - see ProcessMarshalInfo method).

We still need better test coverage for it all - but I'm not ready do it right now. Next step - improve proxy members processing.


iskiselev:
Fixed reading DCE attributes from type proxies. Attributes tests added. Now #731 should be fully resolved.

For now I have no plans to implement full DCE through pried method until backward proxy resolution will be unambiguously (see #769).

iskiselev:
Do you see any problems with it?

Issue #768, opened 2015-07-26
VS2015 issues with ExpressionsTest and ExpressionsExecution


kg:
I was having trouble with the VS2013 service pack installer so I decided to nuke everything and try out VS2015.

Most stuff works, but the Expressions tests fail to run for various reasons (random exceptions, generating bad JS, etc).

This might just be because the BCL has changed but maybe something else is wrong? Haven't been able to make sense of it.

iskiselev:
I've not yet migrated on VS2015 yet. I'll can probably for now install Visual Studio Community Edition in virtual machine to look on it, but I have no Resharper 9. Could you request Resharper license for Open source project for me (https://www.jetbrains.com/buy/opensource/)?

kg:
No time pressure, just wanted to point it out. All the simpler tests seem to work.

kg:
Based on their terms I can't apply for a license for myself or on your behalf. You're welcome to apply for a license as a core committer, I can vouch for you in that aspect.

iskiselev:
Ok, I'll do it myself. I was a little worried with this term:
Site license model
The software may be installed on any number of computers and concurrently used by as many open source project members as the number of licenses provided to the project. 

That's why I supposed that project owner should ask for a licenses. On other side, when I've read it again it looks like core committers can do it themselves.

dmirmilshteyn:
Are the dynamic tests working? On my end, every function that uses dynamic fails to compile. The same code works fine on VS2013.

Even this fails:
```csharp
        public static void Test() {
            dynamic test = new Object();

            test.ToString();
        }
```

kg:
Dynamic tests work here in VS2015 for me, yeah.

kg:
The lvalue branch currently passes everything except for ExpressionsExecution. It looks like the EE failure is due to something outside the compiler.

```
JSIL.Tests.JavaScriptEvaluatorException : JavaScript interpreter exited with code 3 after throwing 1 exception(s):
JSIL.Tests.JavaScriptException: JavaScript Exception: 
Error: No static method with signature 'System.Attribute GetCustomAttribute$588,27=881(System.Reflection.MemberInfo, System.Type)' found in context '<System.Attribute Public Interface>'


JSIL.RuntimeErrorFormat@C:\Users\Katelyn\Documents\Projects\JSIL\Libraries\JSIL.Core.js:10481:9
[...]

PR #771, opened 2015-07-30
.Net Framework 4.6 compatibility fix

iskiselev:
I've temporary excluded System.Resources.FastResourceComparer from translation.
Really we have an bug, that I'll report later. In:
```
    internal static unsafe int CompareOrdinal(byte* a, int byteLen, string b)
    {
      int num1 = 0;
      int num2 = 0;
      int num3 = byteLen >> 1;
      if (num3 > b.Length)
        num3 = b.Length;
      while (num2 < num3 && num1 == 0)
        num1 = (int) (ushort) ((int) *a++ | (int) *a++ << 8) - (int) b[num2++];
      if (num1 != 0)
        return num1;
      return byteLen - b.Length * 2;
    }
```
translation of `*a++` produce `a++.get()` which is syntax error in JS strict mode.

### DIFF
--- a/Proxies/Net46.cs
+++ b/Proxies/Net46.cs
@@ -1,0 +1,11 @@
+﻿namespace JSIL.Proxies
+{
+    using JSIL.Meta;
+    using JSIL.Proxy;
+
+    [JSProxy("System.Resources.FastResourceComparer")]
+    [JSStubOnly]
+    public class System_Resources_FastResourceComparer
+    {
+    }
+}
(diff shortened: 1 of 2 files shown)


iskiselev:
This should resolve #768.

kg:
Hm. I tried adding a test for CompareOrdinal but it couldn't reproduce that translation bug :(

iskiselev:
Will try to reproduce find root cause tomorrow.

kg:
No rush. Just wanted to mention that.

Issue #349, opened 2014-03-22
JSIL.GetType returns type based on contents with numeric variables


griestopf:
The following C# code
float f1 = 3;
float f2 = 3.1415f;
Type t1 = f1.GetType();
Type t2 = f2.GetType();
will result in:
t1 contains System.Single
t2 contains System.Single

JSIL compiled code will result in:
t1 contains System.Int32
t2 contains System.Double

The reason is that JSIL.GetType analyzes the variable contents. Probably hard to solve.

PR #773, opened 2015-08-01
JS Numeric types are wrapped if passed as object.

iskiselev:
I've tried to implement fix for #349 - as was discussed, boxed numeric types are also boxed from Raw JS types in JavaScript.
This PR is still **not ready for merging**, but @kg, I'd like to discuss it with you and probably will need your help.
- With this change dynamic work with numeric types was broken. Please look on DynamicBinaryOperators test for example. As dynamic variables are objects, integers are wrapped. I've not found how can I avoid this behavior, as I have no information that I'm processing translated dynamic CallSite in my WrapLiterals visitors - that information was stripped on function translation
- Second problem is VerbatimVariablesExistingArray test. Integers inside `object[]` are wrapped, so probable `Verbatim.Expression` should unwrap them. On other side, it may not be a problem at all as can be solved inside expression by end user.

Then, not all problems are solved:
- We still need resolve type based on it's information in translation-time if it is unwrapped - as you suggested in first comment to #349.
- Type of expressions like `(byte)12` incorrectly treated as `Int32` - this information is lost inside `ILBlockTranslator.Translate_Box` method. Should discuss how we can solve it.

### DIFF
--- a/JSIL/AST/JSLiteralTypes.cs
+++ b/JSIL/AST/JSLiteralTypes.cs
@@ -364,4 +364,26 @@ namespace JSIL.Ast {
             return PointerType;
         }
     }
+
+    public class JSWrappedLiteral : JSExpression
+    {
+        public JSWrappedLiteral(JSLiteral literal, JSType originalType) : base(literal, originalType)
+        {
+        }
+
+        public JSLiteral Literal
+        {
+            get { return (JSLiteral)Values[0]; }
+        }
+
+        public JSType OriginalType
+        {
+            get { return (JSType)Values[1]; }
+        }
+
+        public override TypeReference GetActualType(TypeSystem typeSystem)
+        {
+            return Literal.GetActualType(typeSystem);
+        }
+    }
 }
--- a/JSIL/FunctionTransformPipeline.cs
+++ b/JSIL/FunctionTransformPipeline.cs
@@ -259,6 +259,8 @@ namespace JSIL.Internal {
 
             // HACK: Something about nullables is broken so we have to do this twice. WTF?
             Enqueue(ReplaceMethodCalls);
+
+            Enqueue(WrapLiterals);
         }
 
 
@@ -505,6 +507,13 @@ namespace JSIL.Internal {
                 ));
             else
                 return true;
+        }
+
+        private bool WrapLiterals()
+        {
+            new WrapLiterals(TypeSystem).Visit(Function);
+
+            return true;
         }
     }
 
--- a/JSIL/JSIL.csproj
+++ b/JSIL/JSIL.csproj
@@ -92,6 +92,7 @@
     <Compile Include="Transforms\CacheTypeExpressions.cs" />
     <Compile Include="Transforms\CacheSignatures.cs" />
     <Compile Include="Transforms\CacheBaseMethods.cs" />
+    <Compile Include="Transforms\WrapLiterals.cs" />
     <Compile Include="Transforms\SynthesizePropertySetterReturnValues.cs" />
     <Compile Include="Transforms\IntroducePackedArrays.cs" />
     <Compile Include="Transforms\StaticAnalysis\AllocationHoisting.cs" />
--- a/JSIL/JavascriptAstEmitter.cs
+++ b/JSIL/JavascriptAstEmitter.cs
@@ -958,6 +958,16 @@ namespace JSIL {
             }
         }
 
+        public void VisitNode(JSWrappedLiteral wrappedLiteral)
+        {
+            Output.WriteRaw("JSIL.Wrap");
+            Output.LPar();
+            Visit(wrappedLiteral.OriginalType);
+            Output.Comma();
+            Visit(wrappedLiteral.Literal);
+            Output.RPar();
+        }
+
         public void VisitNode (JSBooleanLiteral b) {
             Output.Value(b.Value);
         }
--- a/Libraries/JSIL.Bootstrap.Text.js
+++ b/Libraries/JSIL.Bootstrap.Text.js
@@ -550,7 +550,7 @@ JSIL.ImplementExternals(
     $.Method({Static:true , Public:true }, "Format", 
       new JSIL.MethodSignature($jsilcore.TypeRef("System.String"), [$jsilcore.TypeRef("System.Array") /* AnyType[] */ ], []),
       function (format) {
-        format = String(format);
+        format = String(JSIL.UnWrap(format));
 
         if (arguments.length === 1)
           return format;
@@ -756,7 +756,7 @@ JSIL.ConcatString = function (/* ...values */) {
     result = String(arguments[0]);
 
   for (var i = 1, l = arguments.length; i < l; i++) {
-    var arg = arguments[i];
+    var arg = JSIL.UnWrap(arguments[i]);
     if (arg === null)
       ;
     else if (typeof (arg) === "string")
--- a/Libraries/JSIL.Bootstrap.js
+++ b/Libraries/JSIL.Bootstrap.js
@@ -86,7 +86,7 @@ $jsilcore.$MakeParseExternals = function ($, type, parse, tryParse) {
 JSIL.ImplementExternals(
   "System.Boolean", function ($) {
     $.RawMethod(true, "CheckType", function (value) {
-      return (value === false) || (value === true);
+      return ((value === false) || (value === true)) || (value !== null && value.__ThisTypeId__ == $jsilcore.System.Boolean.__TypeId__);
     });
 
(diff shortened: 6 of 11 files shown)


iskiselev:
Also this PR will help us with calling numeric types interface methods.

kg:
Thanks for doing this. It looks like a good starting point. I'd like to rename ```Wrap``` to ```Box``` and ```UnWrap``` to ```UnBox```. That will make it clear what's intended to happen here.

Ideally to match the semantics of .NET boxing/unboxing, we just introduce the idea of a box with a given type, and it becomes possible to box/unbox any given value type at runtime. (As an optimization, boxing a struct could just yield the struct, since structs are already references in JS. But we'd need to make sure it does a copy.) So instead of doing ```new T()``` and setting a special ```__Value__``` property (from a JS perspective there are problems with this), we'd do roughly ```new BoxedValue<T>(v)```.

In this model, ```T.$Cast(b)``` where ```b``` is a boxed ```T``` would just work, so in many cases you could use that in order to handle ```object``` arguments. In cases where you have an object you want to unbox to a specific type without doing value conversions, I think we'd want to introduce ```T.$Unbox(b)``` which does a fast typecheck on the box and extracts a value from it. This could properly handle cases where no box is being used (like structs).

I'm... not sure ```JSIL.UnWrap``` should exist at all. I think they indicate the code in question is doing something it shouldn't, or doesn't have a clear sense of what it actually wants out of the type system.

iskiselev:
I'm not sure that we really want mix .Net boxing and Raw JS-types wrapping. They looks very same, but their usage is different.
We don't want wrap any structure type - if we already created on JS side custom type for structure, it is fully enough and should not be wrapped (at least I don't see why may we need it).
At the same type, we may want wrap non-structure types in future - such as `string` and arrays (as otherwise we lost their type info).
Fortunately, all struct types that we want to wrap are immutable.
I don't understand what you mean when write `new BoxedValue<T>(v)` in JS-side.

If we will introduce some new special JS-side from for wrapped JS-types, it will not resolve one of this PR task - we need some solution to simplify call of interface methods on such structures (or I just don't understand how it will work :) ).

I'm fully agree that each `JSIL.UnWrap` usage indicates some issue in our code, but it is helpful now where we do something bypassing our type system (as in `Format` function in JSIL.Bootstrap.Text.js) or when method should be able to work with both forms of type (implementation of Int32/Float/Char/etc methods). Probable we need find some other solution and get rid of `JSIL.UnWrap`.

iskiselev:
@kg, are you suggesting implement boxing for any structure type when we see `box` il-command?

kg:
Right now we ignore box/unbox operations, which is a mistake. I think it should be possible to replace every case where we need to identify the type of a value at runtime with boxing or compile-time type information. Overload resolution is solved by storing the compile-time type info, for example. The other piece of the puzzle is type information for local types. But I think that's fixable too.

So I believe if we implemented ```box``` correctly such that any time a primitive is converted to ```object```, it gets boxed... that should fix almost everything in this area we care about. The rest of the issues are probably JSIL bugs that change .NET semantics.

I need to think more about this though.

iskiselev:
In .Net, boxing is operation, that converts value-type to reference-type.
On JS side all types are already reference and value-type semantic is emulated. So, we don't need convert anything explicitly (on other side, it may simplify struct-emulation logic and resolve some problems in it).
Right now we have problems with type detection (and method invocations) on small number of types, that we store in native JS-types: ? T[] ?, ? string ?, bool, char, sbyte, byte, short, ushort, int, uint, float, double, decimal.
I'm pretty sure that local variable type resolution could be simple resolved with proxies (will try to do it to prove).

If you want to translate all box/unbox, this logic should be moved from FunctionTransformPipeline step to IlBlockTranslator. I afraid to do it, as it may break sensitive struct-emulation logic and a lot of other pipelines steps, that I don't know good enough. :(

#349 that I've tried to solve here is blocking me of moving from JSIL BCL class implementations to fully translated BCL with preserving only minimal subset of JS classes inside Libraries folder. Ideally, their will be special parameter - UseJsilBclImplementation, that will prevent (if false) from loading most JSIL.Bootstrap.* (some of JSIL.Bootstrap.* files should be split for it).

kg:
Can you call out some specific BCL methods affected by this (or provide them as tests) so I can make sure I fix it one way or another?

iskiselev:
Expression.Constant(0d, typeof(double))

iskiselev:
Looks like I once again partially blocked by this problem. Now I have problem when try cast Numeric type to IConvertible.
Even when I implemented IConvertible members, cast/is operation will still fail - as raw JS type has no `__ThisTypeId__`.
Let's discuss once more Boxing/Wrapping solutions and problems that they solve:
**Wrap**
- Applicable only to types (both struct/classes) that translated as JS native types (Numbers, String/Char, array?)
- Wrapped when source is passed as object/interface type, story original type information

**Box**
- Applicable to all structs when they passed as object/interface
- Store information about source object type. Do we really need it for non-raw JS objects? What problems will it solve?
- How it will work for generics with type constraint (as there will be no box command in IL), i.e.:
```
using System;

public static class Program
{
    public unsafe static void Main(string[] args)
    { 
        Console.WriteLine(Sampe(10));
    }

    private static short Sampe<T>(T input) where T : IConvertible
    {
        return input.ToInt16(null);
    }
}
```

iskiselev:
At the same time, using `new BoxedValue<T>(v)` in JS instead of Value field  may work in Wrap solution also.

kg:
I think we only care about types that are translated as JS native types. So: numbers, char, string, array. We need to maintain the JS protocols, though, which is a real problem. I think our only choice there is to not wrap strings or arrays. For arrays, we can probably attach type information to them with a weakmap or a custom named property (the latter works alright in engines now, I believe). For strings I think we will stick with JS String and just have to deal with the fact that they don't ever have referential equality. (Though maybe ```str[i]``` works on ```new String``` in JS? I've never tried.)

So I think the path forward looks mostly like what you have here with wrapping. We have a generic 'box' type, maybe just an improvement on the one currently used for ref/out vars, and it is used to store boxed numeric types (including ```char```). We switch char over to a numeric type.

The box type will need to get smarter: ```toString``` and ```valueOf``` on it will need to be implemented appropriately, such that ```'hello' + (new Box<char>(33))``` produces ```'hello!'```. I think that's probably not too hard to do, though. Most (all?) of the code JSILc generates won't rely on that property, but we want to make sure JS interop scenarios don't end up totally hosed. There will definitely be issues where JS grabs an unboxed numeric type out of a field and passes it back into C#, but I think that will just be a place where JS needs to obey protocols when dealing with JSIL data structures (maybe in special cases we can have an attribute that forces fields to always be boxed so it's safe for JS to touch them.)

iskiselev:
@kg, looks like we already have "JSIL.BoxedVariable", but I'm not fully understand when and what for we use it.

kg:
It's a ref cell for ref/out parameters.

iskiselev:
I'm working on new version based on our discussion. Looks like I will complete it today. One still open question is performance. Here is UnsafeIntPerformanceComparison test. As `TestInlineAccess` use Verbatium, int passed as argument was boxed (it could be easily fixed with creating several generic Verbatium methods with 1-10 args). Let's look on performance in Chrome (in IE and Firefox it is nearly same):
Original version:
```
    var pBuffer = JSIL.PinAndGetPointer($thisType.Buffer);
    var offsetInElements = pBuffer.offsetInBytes >>> pBuffer.shift;

    for (var i = 0; i < 1048576; i = ((i + 1) | 0)) {
      pBuffer.view[(offsetInElements + i) | 0] = (i % 255) | 0;
    }
```
Box version:
```
    var pBuffer = JSIL.PinAndGetPointer($thisType.Buffer);
    var offsetInElements = pBuffer.offsetInBytes >>> pBuffer.shift;

    for (var i = 0; i < 1048576; i = ((i + 1) | 0)) {
      pBuffer.view[(offsetInElements + $T00().$Box(i)) | 0] = ($T00().$Box(i) % 255) | 0;
    }
```

Original version:  00505.00ms
Box version: 186131.00ms

It's awful. According to profile, all time was spent for only arithmetic operation - $Box and valueOf has taken very small part of it.

And now strange thing. If I put breakpoint in this function before cycle and execute same cycle from Developer console, it takes about 2000ms for original version and 12000ms for box version. Looks like there are some problems inside optimization :(

May be it is not very big problem - as we should not use boxed form a lot, but it really strange.

kg:
It's fine if TestInlineAccess isn't fast anymore, it's just a hack. That is
a tricky downside to boxing though, and could affect other
performance-sensitive code.

iskiselev:
I'm much more curious why it works 15 times faster from developer console. It really looks like a bug in JavaScript optimizer. Probably we need to file some tickets :)

Issue #777, opened 2015-08-03
Option to run C#/VB test cases with Roslyn


iskiselev:
As I understand, currently all test cases are compiled with pre-Roslyn compiler.
As Roslyn use different patterns, we should verify that JSIL is still compatible with it.
Looks like Roslyn-based CodeProiders are inside [nuGet packege](https://www.nuget.org/packages/Microsoft.CodeDom.Providers.DotNetCompilerPlatform/). I was unable two run with it, but I spent very small time for it.

Also good idea would be introducing options to compile with/without optimizations (Release/Debug), as this modes also use different patterns.

Issue #776, opened 2015-08-03
Roslyn null check translation


iskiselev:
JSIL incorrectly translate Roslyn compiled (debug) comparison with null.
Here is a test case:
```
using System;

public static class Program
{
    public static void Main()
    {
        Run(new Derived());
    }

    public static void Run(Base b)
    {
        var a = b as Derived;
        if (a != null)
        {
            a.Method();
        }
    }
}

public class Base { }
public class Derived : Base {
    public void Method()
    {
        Console.WriteLine("Method");
    }
}
```

Expected result: "Method".
Actual result: Nothing.

Method `Run` in CIL:
```
	IL_0000: nop
	IL_0001: ldarg.0
	IL_0002: isinst Derived
	IL_0007: stloc.0
	IL_0008: ldloc.0
	IL_0009: ldnull
	IL_000a: cgt.un
	IL_000c: stloc.1
	IL_000d: ldloc.1
	IL_000e: brfalse.s IL_0019

	IL_0010: nop
	IL_0011: ldloc.0
	IL_0012: callvirt instance void Derived::Method()
	IL_0017: nop
	IL_0018: nop

	IL_0019: ret
```

It was translated to:
```
    var derived = $T00().$As(b);
    if (derived > null) {
      derived.Method();
    }
```

So, Roslyn compiles comparison with null `cgt.un` instead of `ceq` and this should be translated as special case, as in JavaScript `object > null = false`.

PR #778, opened 2015-08-03
Fixed Roslyn null comparison.

iskiselev:
Should solve some more problems with #776, #768.
We still need to add some tests (see #777). Right now it may be a good idea to add Binary Test.

### DIFF
--- a/JSIL/ILBlockTranslator.cs
+++ b/JSIL/ILBlockTranslator.cs
@@ -1749,6 +1749,23 @@ namespace JSIL {
                     }
                 }
             }
+            else if (
+                (node.Arguments[0].ExpectedType == node.Arguments[1].ExpectedType) &&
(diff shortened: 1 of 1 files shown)


JSIL/ILBlockTranslator.cs
```diff
@@ -1749,6 +1749,23 @@ bool forDynamic
                     }
                 }
             }
+            else if (
```
kg: This is a little awkward but I'm willing to merge it.

An explanatory comment would be great, and maybe reduce the redundancy of the if conditions. I think most of the stuff in the 'else if' condition here is redundant with the conditions inside of that else if block.

Both of those improvements can come later though.

kg: Also, I seem to remember ILBlockTranslator already having some fairly general logic for dealing with goofy comparisons/arithmetic on nulls, so if this could sit next to that existing stuff or replace it, that would be awesome.

Issue #779, opened 2015-08-04
JavaScript libraries processing


iskiselev:
While I'm working on replacing JSIL .Net implementation from Libraries folder, I understand that I want to introduce some pipeline for processing .js Scripts before they go to Libraries folder.

So, on first step, I'd like to split JSIL.Bootstrap.*.js to some smaller files (probably per-call basis).
Then, I need introduce two JS configurations - JSIL implementations and BCL implementation. Only some of per-class .js files will be included in resulting .js files for BCL implementation, as opposite to JSIL implementation - they should include all content and looks same as they are now. Later we probably need some more preprocessing logic.

As [Visual Studio now support using grunt/gulp tasks](http://www.hanselman.com/blog/IntroducingGulpGruntBowerAndNpmSupportForVisualStudio.aspx), I want to create new project, that will contain all required tasks with grunt that will be executed on Solution Build.

Cons of this solution - as I understand, user who tries to Build JSIL in this case should have Node.JS installed (probably not required for VS 2015). I'm not sure if this will work in VS 2012 and prior versions, same as in Express edition. It could be solved partially by providing .cmd file that will be still able to build it all - if user has Node.JS installed.

Initially original JSIL.Bootstrap.*.js files will be created from per-class split .*js files using `grunt-contrib-concat` task, then, if needed we may move to `grunt-preprocessor` or `grunt-preprocess`.

@kg, what do you think about it? Can you suggest some other options? I really need to split JSIL.Bootstrap.*.js, but still want to be in sync with master :)

kg:
I'm not OK with requiring node or anything heavyweight like that. If it's really necessary I'm OK with having a build step of some sort, though, especially if it gets us towards always using the real BCL instead of those annoying hand-written stubs.

iskiselev:
What if we will provide .cmd, that will install silently node.js locally to JSIL folder? Or you don't like idea of dependency to node.js at all?

Honesty, right now I'm able to concat several *.js files into one with only msbuild, but I've played a little bit with it - and can say now that with node.js infrastructure it will be most convenient. Also, looks like Microsoft also advice use node.js infrastructure for such tasks - that why node.js support is built-in in ASP.NET 5.

Another option will be preserve *.js in Libraries folder - so most end users will not need node.js at all. I don't like this option, as I preffer be sure that all changes are done in original *.js and not that were processed and placed in Libraries - and best option for it will be just ignore Libraries folder at all.

kg:
[...]

PR #780, opened 2015-08-05
Js preprocessing

iskiselev:
Here is example of what I want for #779.
There is no any changes to any user in compilation process - you need just check out and build.

Inside JSIL solution was added new projects - JSIL.Libraries. All files from Libraries folder moved to this new project.

Then I was able to start splitting JSIL.Bootstrap.js to much smaller files. All parts extracted from it went into JSIL.Libraries\Includes folder - currently as an example I extracted very few classes, but will continue this process if we agree.

So, now in you can see this JSIL.Bootstrap.js block:
```
// #include "Bootstrap/Core/System.String.js"
// #include "Bootstrap/Core/System.Exception.js"
// #include "Bootstrap/Core/System.Boolean.js"
// #include "Bootstrap/Core/System.Char.js"
// #include "Bootstrap/Core/System.Byte.js"
// #include "Bootstrap/Core/System.SByte.js"
// #include "Bootstrap/Core/System.UInt16.js"
// #include "Bootstrap/Core/System.Int16.js"
```

This will allow me simply create something like BCL\JSIL.Bootstrap.js, that will contain only minimal amount of classes/functions. It is really easy now remove only parts of `JSIL.Bootstrap.js` and check how it works. 

Besides, as there is also `#include_once` directive, we could easily manage utility functions - as an example, I've extracted some in Bootstrap/Core/Utils folder (look how System.Int16.js, Bootstrap/Core/System.UInt16.js and Bootstrap/Core/System.Boolean.js use them).

@kg, this change is already good for merge. So, I'd like to discuss it ASAP, as otherwise I'll blocked on my tasks of moving to full BCL. Thank you.


iskiselev:
Additional note: if solution is opened in VS 2015, you can open View->Other Windows->Task Runner Explorer and execute task to build *.js files from this window.
Same is possible with VS 2013 if Task Runner Explorer extension is installed.

iskiselev:
File, that describe required tasks is `gruntfile.js` (grunt-preprocessor was used there).

iskiselev:
If later you will find any problems with this solution, it could be very simple reverted - you just need build it once, make Libraries folder non-ignorable, delete JSIL.Libraries project and commit back generated Libraries.

kg:
Some thoughts from a brief skimming:

Splitting the individual types out into files in Sources/ seems reasonable. Having a .csproj that 'owns' the output JS is a good way to set up build dependencies and other things, since that way an application can pull it in as a dependency along with JSIL.Meta. In theory, anyway. The msbuild integration seems straightforward as well.

On the other hand... this is like 6 nuget packages to do an ```#include``` preprocessing on the .js file. This just feels too heavy for a relatively small thing. I also don't want us to check ```node.exe``` into the tree - I did this in the past with some dependencies and it was a huge mistake we shouldn't repeat. (I'd actually like to purge all the binaries from the tree with a history rewrite and move them to a submodule. I'm not sure how feasible that is.)

If we can do this in a less heavyweight manner I think I'd otherwise be happy to check it in. I don't like having a build process to generate the .js file but it's probably about time that it's necessary.

iskiselev:
Really we don't need all this .nuget packages.
We really need only: MSBuild.Node, MSBuild.Grunt, MSBuild.Npm.
Also, we can use Node.Js also from Nuget package: Node.js.
So, there will be no binary reference in that case.

Right now I haven't found any lighter way to do this work - at least none of other solutions will have so good ecosystem as Grunt based.

All this packages are really needed only for MsBuild integration. In we will build in Visual Studio 2015 all this tasks will work without any of this Nuget packages at all (unfortunately it will work only inside VS, as integration use VS events to insert this tasks, that is not acceptable for automation builds/CI).

So, right now, I'll continue splitting JSIL.Bootstrap.js - and we could try to find some other way to integrate this tasks in build.

kg:
I guess my confusion is over why we need an ecosystem at all. These are just text includes. Why are grunt, gulp, npm and node needed? I could understand it if something more sophisticated were happening.

iskiselev:
Can you suggest any simpler good-supported system that will be able to process include, include_once and probably define/ifdef for JavaScript files that could be easily integrated with MsBuild?

I spend a day and was unable to found anything like this.

Besides, right now we only use it to preprocess - but I suppose that we could late add minifing/bundling, maybe will use some other modern *.js processing options.

As I can see, main ecosystem for *.js processing is Node.JS and all around it nowadays. Maybe I'm not right :) When I added all this (and besides I still have not found any lighter solution) I decided, that probably this ecosystem is good choice (I agree that we can argue on this!).

So, I will be glad to move to any lighter option (at least for now) - please just suggest it!

iskiselev:
At the end, what is really difference how this tasks are implemented? Will it be special MsBuild task, external .exe file or program that is executed under Node.Js?

Really Node.Js option even has benefit - it is cross-platform, so for we still have option to run this processing on Mac/Linux (probably not so straightforward until somebody will test it/improve it). It may be much harder even with custom MsBuild task.

kg:
In the past I've used ```tcc.exe``` in preprocessor-only mode but it emits #line directives, so that won't do.

I don't do this class of webdev very often. For the short term could we just use a build step? I think anything sophisticated like minifying and bundling should be things JSILc knows about, since end users will want to do them to their compiled apps, not just to the JSIL stdlib.

Would T4 cut it? T4 is super awkward but it has integrated support in VS and can generate pretty much any text programmatically at build time and cache it.

Or, to put it another way: Can we make the stdlibs a manifest just like the manifest produced when JSILc compiles an executable? Then, can we solve all these problems by improving the JSILc ecosystem so manifests are more powerful? Bundle-into-single-file, bundle-into-archive, etc.

I'd be happier doing the work to generalize JSILc's tooling so that it can solve the problems that apply to the stdlib instead of pull in a bunch of dependencies as a one-off.

iskiselev:
Creating of Libraries (at least part of them) may be moved inside JSIL in future, but this task is much harder. Right now I have not enough time to solve it. If we will include an options to bundle/minify inside translation time, it may be very good option for most simple scenarios (not all of them, for example dynamic loading application part will require some other solution on user side).

I believe this could be solved with T4, but good solution, that will generate all files in specific folder (../Libraries for us) will require a lot of code and time to develop.

So it seems that we try here to reinvent the wheel.
Main problem for me right now - as soon as I've done this changes I would be unable to automatically merge with master. That's I try to push this solution or some other solutions, but I really need anything quickly - and unfortunately I already spent too many time for it.

iskiselev:
I've changed PR so that dependency to Node.Js and NPM are restored through NuGet, so we don't have Node.Js binaries in Upstream.
Please update to new version for further review.

kg:
I really strongly dislike the node stuff, but I don't have the time to come up with a better solution to this. I wouldn't want the amount of work you poured into this to go to waste, so I guess I'll merge it... I don't see any problems with the rest of the PR now that node.exe is not checked in.

Thanks for doing the work to split all those files out. It was something I considered doing in the past and it will definitely make future improvements easier.

iskiselev:
I'll really open for future discussion of it. Later I'll have some more time to dig into it - if I'll found an option to remove Node.Js dependency, I'll be happy to do it!

kg:
It sounds like there just needs to be infrastructure so this can be solved with JSILc manifests. Maybe I can find time to build that.

One thing I noticed that you should fix: For some reason grunt is stripping newlines when gluing the files together. So where previously there were line breaks, there aren't linebreaks anymore. Due to ASI this can change how the JavaScript actually parses, and it might introduce bugs.

For example:
```javascript
...

  return function (lhs, rhs) {
    return compare.Call(comparer, null, lhs, rhs);
  };
};﻿JSIL.MakeClass("System.Object", "System.Collections.BitArray", true, [], function ($) {
  $.Property({ Public: true, Static: false }, "Method");
});

...
```

I have historically tried to be vigilant about spraying extra semicolons everywhere, but I doubt I was consistent enough for this to not be a problem.

kg:
Can you file an issue about migrating the libraries to a manifest, and describe the scenarios you care about there? Or at least what manifests would need to be capable of for your use cases.

From glancing over things it seems like all you really need is basic dependency resolution, whether it's done through preprocessing (like now) or dependency loading or what have you. But I don't want to underestimate what you're trying to solve here.

It seems like it might also be desirable to glue compiled C# mscorlib + hand-written JS files together into a single manifest that's loaded as one unit? jsilcorlib, essentially? Is that the sort of thing you have in mind here?

This interacts with the bytecode stuff and related issues I've been considering, so I'll be thinking about this angle...

iskiselev:
I'll create an issue later - I'm nor ready to formalize requirements for built-in manifest processing. Will think a little bit more on it.

Issue #791, opened 2015-08-10
Move NUnit dependency to NuGet


iskiselev:
@kg, as you stated that we should move binary dependencies out of project, should we replace nUnit dll with [NUnit](http://www.nuget.org/packages/NUnit/) and [NUnitTestAdapter](http://www.nuget.org/packages/NUnitTestAdapter/)? It will also should allow everyone run test cases from VS test launcher without any plugin installation.

kg:
Sounds like a good idea. It'd be good to see if there's a way to get vs2015 to do threaded test runs... The test suite keeps getting bigger and slower.

On August 9, 2015 10:23:22 PM PDT, Igor Kiselev <notifications@github.com> wrote:
>@kg, as you stated that we should move binary dependencies out of
>project, should we replace nUnit dll with
>[NUnit](http://www.nuget.org/packages/NUnit/) and
>[NUnitTestAdapter](http://www.nuget.org/packages/NUnitTestAdapter/)? It
>will also should allow everyone run test cases from VS test launcher
>without any plugin installation.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/sq/JSIL/issues/791

- kg (mobile)

PR #792, opened 2015-08-10
NUnit and x64

iskiselev:
NUnit dependency moved to NuGet (#791).
Changed x86 configuration AnyCPU for JSIL, JSIL.DeadCodeAnalyzer, Tests, SimpleTests, Tests.DCE.
I leaved x86 as only configuration for Compiler project to not break compatibility with XNA. For it I moved IAnalyzer interface from Compiler to JSIL project (so JSIL.DeadCodeAnalyzer do not depend on Compiler anymore, same as Tests.DCE).

Really, x86 configuration is critical only for .exe project, as it choose subsystem to run - all .dll projets should be always compiled as AnyCPU unless they do not reference x86 native dlls :).

Now we could run all tests on x64 subsystem: for it in VS menu select Test->TestSettings->Default Processor Architecture->X64.


iskiselev:
As test case may be x86 only, I've changed loading assembly to  AssemblyReflectionOnlyLoadFrom for .exe test cases (and CPP/IL code generators when we generate executable).

Really we should consider option to always load compiled test assemblies in external application, as we now can execute test from same assembly multiple times (and some test case is not safe - each execution modify their static variables state).

Also we should discuss if we can skip loading assemblies at all 0 we really need only file name from it. It is impossible to unload assembly (except if we'll use  different app domains for them). So, it should have good influence on our tests memory usage.

kg:
I'd prefer to load test assemblies in a temporary appdomain, so we don't have to deal with even more ghost processes. But you make good points.

iskiselev:
I also thought today a little bit with AppDomain, but unfortunately AppDomain still may not solve problem of loading x86 app (all EcmaScripten tests, as they reference x86 native .dll) in x64 app.

kg:
Looks good, thanks.

kg:
If I set the solution to 'Any CPU' I get a bunch of errors building after this merge. The other two configurations (x86, .net 4.5 x86) still work.

I'd be OK with pruning all the custom .net 4.5 stuff and just having two configurations for x86 and AnyCPU, if the old setup is getting in your way.

iskiselev:
I've always had question, what for we need custom .net 4.5 configuration? I've seen target, but how does it affect on our build? I haven't seen any #ifdef for 4.5

kg:
It used a custom set of targets files to let you build against the 4.5 framework without having to install a newer build of Visual Studio. I had it because of XNA compatibility issues (ones that have since been resolved), and because 4.5 was still new at the time.

iskiselev:
What error do you have? Are they in XNA-dependent projects? Where I should install XNA now, as I do not build them?

kg:
```ECHO JSIL must be built with the platform set to x86```

But also, yeah, the XNA stuff. I think we just need to set the configuration to not build them.

iskiselev:
As soon as #793 will be resolved we may have only one configuration. All non-xna *.dll projects will be AnyCPU, and two executors in x86 and (x64 or AnyCPU - should be discussed) always.

iskiselev:
Really, we can have two solution configurations: .Net 4.0 and .Net 4.5 if we need it.

Issue #794, opened 2015-08-11
CI for JSIL


iskiselev:
I think it will be good if we'll have CI.
Travis-CI provides free cloud CI for open-source projects. I was able today to build project and run Test cases [there](https://travis-ci.org/iskiselev/JSIL/builds/75019204).

Unfortunately, as they run on mono we was unable to translate mscorlib, but I hope we'll be able to run most other tests.
We still need solve problem with too big output from compile/test executions to make it more usable.

kg:
If it's free we can definitely use it, though if it's gonna be central to the project I might want to run it myself. I have a linux box, but I'd prefer to run the CI stuff on a windows box if possible since that's where most users will be. (Having an additional CI box for linux is a good idea, though...)

iskiselev:
I was able to run tests (on my brunch right now).
Looks like we fail a lot of tests on that box. https://travis-ci.org/iskiselev/JSIL/builds/75038382

iskiselev:
As currently 10 tests are failing on Mono, probably I should fix CI script so that it report successful build if there is no more then 10 broken tests until we fix them?

kg:
Is there a way to just have it report the # of passed/failed tests in the github UI? Or we can put tests known to fail on Linux into a separate category, sort of like FailingTestCases.

iskiselev:
Excluded failing tests from CI (added to #798 PR).
As I understand, for providing additional information (number of passed tests and probably something more) we need use GitHub API from CI script. Will be good to implement :).

iskiselev:
CI sometimes hang. Still don't know which test to blame. Probably we need split tests execution into smaller groups (edit CI config) to detect it.

iskiselev:
I've found it - ThreadingTests.WaitThrowsIfNestedDeadlockWouldHaveOccurred is responsible for often freeze on Mono.

iskiselev:
One more CI (free for open source) - http://www.appveyor.com/. It runs on windows, but have time limit (30 minutes) per build. So I've created config for it #828, that will build JSIL and run tests that don't work on Mono (except emscripten integration).
@kg, if you merge it and activate build there, we will have nearly full test coverage on CI.

PR #798, opened 2015-08-13
Test: AppDomains and memory usage

iskiselev:
Here I moved each test execution to separate app domain which is unloaded after with test assembly. So, test execution no more affect on each other if classes have static states.
Also I fixed work with string on writing output to file, so there is no more OutOfMemory exception even in translated tests on 32 bit.

Fixed some small EOL and added them to .gitattributes/.hgeol.

Unfortunately, with introducing app domains Simple test cases executes nearly twice as long then before.

Probably we could find some options to run most test cases in same domain, it should be able to change in just one line.


Tests/ComparisonTest.cs
```diff
@@ -160,30 +160,49 @@ public static string EvaluatorPrepareEnvironmentCode(Dictionary<string, string>
...
+
             switch (extensions[0]) {
                 case ".exe":
                 case ".dll":
                     var fns = absoluteFilenames.ToArray();
                     if (fns.Length > 1)
                         throw new InvalidOperationException("Multiple binary assemblies provided.");
-
-                    Assembly = extensions[0] == ".exe" ? Assembly.ReflectionOnlyLoadFrom(fns[0]) : Assembly.LoadFrom(fns[0]);
+                    AssemblyUtility = CrossDomainHelper.CreateFromAssemblyPathOnRemoteDomain(AssemblyAppDomain, fns[0], extensions[0] == ".exe").AssemblyUtility;
```
iskiselev: If we pass here AppDomain.CurrentDomain, it will work in same domain. We could introduce some options later to make it run faster (but with more memory usage/state effects).

Tests/ComparisonTest.cs
```diff
@@ -160,30 +160,49 @@ public static string EvaluatorPrepareEnvironmentCode(Dictionary<string, string>
 
             JSFilenames = null;
 
+            AssemblyAppDomain = AppDomain.CreateDomain("TestAssemblyDomain", null, new AppDomainSetup
```
iskiselev: If we pass here AppDomain.CurrentDomain, it will work in same domain. We could introduce some options later to make it run faster (but with more memory usage/state effects).

kg: We could probably do what the runner does with js.exe and use a thread to spin up appdomains in advance, assuming that appdomains aren't stop-the-world.

It would probably also be okay to just cache the output of the C# tests, in which case we don't need to run them if they haven't changed. That might be a worthwhile step at this point.

iskiselev:
Travis really start working! Cool! Unfortunately for now we see false-broken message, as 10 test cases still not work correctly on Mono.

iskiselev:
I really like your idea of result caching. In that case we probably will not need app domains at all (I believe it is not so critical to unload assembly :) ).
Pre-creating app-domain with multiple thread is also worth to try - but for now I'd like to return to my main task with moving to translated BCL.

iskiselev:
So, with this fix I've seen all tests (except one broken) passed in VS Test launcher from first time in 32 bit configuration.

kg:
Don't worry about appdomains for now, but it's good to have the option. Result caching is probably what I'll go for. Does need a fallback to actually run the test any time a debugger is attached (for cases where you're trying to debug the C# test), but that's not too hard.

iskiselev:
Per discussion in #794, moved all simple tests failing on mono to separate folder (SimpleTestCasesFailingOnMono) - so that they have separate category. All this tests are excluded from CI run.

iskiselev:
@kg, will you merge it or do you want move out some parts of this PR (I can split parts if need)?
I've continued work on enabling test on CI, their compatibility and marking tests as FailsOnMono, so I'm not sure what commit select as base (as I move some tests to sub-folders I'd like to know it for minimal merge).

iskiselev:
All tests (except translated bcl and ecmascripten) runs on Travis-CI. Tests that fails on Travis-CI marked with FailsOnMono/FailsOnMonoWhenStubbed category and are not executed.
Compatibility of some tests improved.
So here is some statistics:
Tests marked with FailsOnMono: 10*2 (Simple) + 6 (Other in subfolders) + 19 (Marked with attribute) = 45
Tests marked with FailsOnMonoWhenStubbed : 2

Total test passed on Travis-CI: 1226
Total test ignored on Travis-CI: 4

kg:
One of the recent changes (probably in this PR)? broke unit tests for me somehow. My first attempt to run tests hung forever, and now unit test discovery fails with some guid errors. It seems like maybe having the test adapter as an nunit package in the project is incompatible with installing the adapter into VS's extensions gallery. Personally I don't think it should be in the project's nunit packages, is there a reason why you put it in there?

kg:
```
NUnit VS Adapter 2.0.0.0 executing tests is started
Loading tests from C:\Users\Katelyn\Documents\Projects\JSIL\bin\SimpleTests.dll
Run started: C:\Users\Katelyn\Documents\Projects\JSIL\bin\SimpleTests.dll
// BclBootstrapStubbed.cs ... 
generated
```

It hangs here every time. If I attach a debugger when running tests it seems to proceed without any issues. I'm not sure how to debug this :/

kg:
The problem appears to be that test runs can leak threadpool workers waiting on stdout/stderr. They never wake up because of how we've torn down the appdomain (and left js.exe alive?) so eventually the threadpool is full of dead threads. For some reason this makes ComparisonTest.Dispose hang with 0% CPU usage. Not sure about that one.

iskiselev:
Have never seen both of this problems. Strange. Will try to install nunit test adapter from extensions gallery - probably it is key element here?
We could uninstall package with NUnit VS test adapter if you want or if we found that it brokes something - right now I installed so that everyone could run unit test without additional steps.

kg:
No matter what I do, AppDomain.Unload hangs indefinitely. :-(

iskiselev:
Are you talking about AppDomain.Unload introduced in this PR (AppDomain that used for compiling and executing tests)?

iskiselev:
Probably we lock Console in some multi-threading scenarios?

iskiselev:
Do you have same problem when run from nunit gui/console test runner?

kg:
As far as I can tell though, this means we're disposing the appdomain while the test is still running? I can't figure out why that would be the case.

iskiselev:
Can you try change lines that create AppDomain and test with using AppDomain.CurrentDomain instead?

kg:
I'm seeing some things to the effect that AppDomain.Unload is intrinsically unsafe in nunit but supposedly it was fixed? Maybe not in the version we're using...

iskiselev:
Why I don't see this problems neither there is no problems on Travis-CI?
Can you try to uninstall VS plugin for nunit and check without it?

When do you see this problem? After each test or only after some specific?

kg:
I uninstalled the VS plugin already, this didn't work at all with it installed.

It looks like only the bcl bootstrap test run that happens in the fixture setup is the one that hangs. Maybe it's some sort of deadlock given the huge amount of stdout spew?

kg:
This only hangs if I click 'Run All' in the VS test runner, too. If I just pick a few tests from the stubbed bcl fixture and run those individually they work.

kg:
After the AppDomain.Unload hangs, ~20 pages of stdout spew from JSIL gets sprayed to the test output window. Is it something about how you deal with the output from that test compile, maybe?

iskiselev:
Do you have same issue with Translated fixture setup?

Really there is no output to console in Bootstrap at all (in that app domain). Can you try add Console.WriteLine("a") to both bootstraps?

kg:
The console output is the error messages from compiling mscorlib, about untranslatable pointer expressions etc.

iskiselev:
So it is output from translation. It happens in main app domain (but it still lock Console) Hmmm...

kg:
I'm disabling appdomains for now. It seems like this is probably due to how output redirection works in the VS test runner.

kg:
(Ugh, hit enter too early)
In the future if we want to turn AppDomains back on, this will fail in an understandable way because I put a timeout on the appdomain unload operation.

iskiselev:
Ok, cool! The most strange here is that I also test it under VS Test Runner (same as Resharper test runner also).

iskiselev:
Let's do UseAppDomains based on environment variable to not break CI?

kg:
Feel free to do that if you think it's a good idea. Why does CI need AppDomains?

iskiselev:
One more question - which test case VS Test runner executes first on your machine? On my - it is SimpleTests from SimpleTestCasesForStubbedBcl. Probably that's why I don't see any problems (as fixture is executed before any other tests).

iskiselev:
CI needs AppDomains as otherwise tests that modifies static field will not pass on second execution in same session (Simple or Stubbed).

iskiselev:
Hm. I could split Simple and Stubbed tests in to different nunit test runners, so there will be no problems - it is another option.

kg:
For me I'm pretty sure the stubbed bcl simple test cases were always running first, and hanging in appdomain unload.

iskiselev:
So no tests are passed at this case?

kg:
Yeah, it was hanging on the first test run in the setup of the fixture.

FYI With appdomains disabled, everything is working except that translated BCL test setup is running out of memory again. I probably need to switch to the 64-bit test runner.

iskiselev:
I've tested this today on another computer (VS 2013 Test runner) and still was unable to reproduce hang :(.
I've checked - really the only console output of stbbed bootstrap:
```
// SimpleTestCasesForStubbedBcl.SetupFixture() {
// BclBootstrapStubbed.cs ... 
generated
// SimpleTestCasesForStubbedBcl.SetupFixture() }
```

So, it really small (it is not big output of translated tests bootstrap). I still have no idea what have happened.

iskiselev:
Can we just ignore not-unloaded domain? Could you check what will happen if we just warn about it, but continue execution (will it block further domains from unloading)?

kg:
It doesn't produce output when it works correctly, only when it falls, which is mysterious

On August 14, 2015 10:29:36 PM PDT, Igor Kiselev <notifications@github.com> wrote:
>I've tested this today on another computer (VS 2013 Test runner) and
>still was unable to reproduce hang :(.
>I've checked - really the only console output of stbbed bootstrap:
>```
>// SimpleTestCasesForStubbedBcl.SetupFixture() {
>// BclBootstrapStubbed.cs ... 
>generated
>// SimpleTestCasesForStubbedBcl.SetupFixture() }
>```
>
>So, it really small (it is not big output of translated tests
>bootstrap). I still have no idea what have happened.
>
>---
>Reply to this email directly or view it on GitHub:
>https://github.com/sq/JSIL/pull/798#issuecomment-131300817

- kg (mobile)

iskiselev:
It is really interesting. Could you provide this output? Maybe it will give me some ideas?

kg:
How hard would it be to set things up such that SimpleTests / Stubbed BCL / Translated BCL each run in a single shared domain for all the tests? I'm not sure that's compatible with how things were set up, but that would avoid most of the domain setup/teardown issues, and fix the problems with those tests sharing statics between each other.

iskiselev:
Should be not very hard - we just need path AppDomain to ComparisonTest constructor from outside, but it still may increase memory usage so that some tests will not work on x86.
Really solution with saving test output will be much more elegant and should not be much harder. I still keep it in mind, but right now have some tasks with higher priority.

kg:
I may do the saved output thing myself.

PR #801, opened 2015-08-15
More CI tests

iskiselev:
Some more tests fixed to be compatible with mono. AppDomains in test could be activated by adding non-empty "JsilUseAppDomainsInTest" environment variable (enabled on CI).
Now we execute 1238 tests (+12).


iskiselev:
@kg, are there anything in this PR that should be discussed? Maybe I should extract "	UseAppDomains based on JsilUseAppDomainsInTest environment value" and "Marked ThreadingTests.WaitThrowsIfNestedDeadlockWouldHaveOccurred FailsOnMono, as it often freeze." commits - as with them we will have trusted results from Travis-CI?

kg:
Hadn't looked at it, thanks for the ping

Issue #806, opened 2015-08-20
Incorrect method override selected for delegate call


iskiselev:
Incorrect method override selected for delegate call. Look next test case:
```
using System;

public static class Program
{
    public static void Main(string[] args)
    {
        Action<Derived> dDerived = Method;
        Action<Base> dBase = Method;
        Action<IInterface> dIInterface = Method;

        var obj = new Derived();
        dIInterface(obj);
        dBase(obj);
        dDerived(obj);
    }

    public static void Method(Base input)
    {
        Console.WriteLine("Base");
    }

    public static void Method(IInterface input)
    {
        Console.WriteLine("IInterface");
    }

    public static void Method(Derived input)
    {
        Console.WriteLine("Derived");
    }
}


public class Base : IInterface
{
}

public class Derived : Base
{
}

public interface IInterface
{
}
```

Expected result:
```
IInterface
Base
Derived
```

Actual result:
```
Base
Base
Base
```

PR #807, opened 2015-08-21
Invocation overloaded method through delegate.

iskiselev:
Here is fix for #806 and test for it.

### DIFF
--- a/JSIL/SpecialIdentifiers.cs
+++ b/JSIL/SpecialIdentifiers.cs
@@ -111,18 +111,54 @@ namespace JSIL {
             return new JSNewArrayExpression(elementType, dimensions, initializer);
         }
 
-        public JSInvocationExpression NewDelegate (TypeReference delegateType, JSExpression thisReference, JSExpression targetMethod, JSMethod method) {
+        public JSInvocationExpression NewDelegate (TypeReference delegateType, JSExpression thisReference, JSExpression targetMethod)
+        {
+            var targetDotExpression = targetMethod as JSDotExpressionBase;
+            var jsMethod = targetDotExpression != null ? targetDotExpression.Member as JSMethod : null;
+
+            JSExpression[] invocationExpressionArguments;
+
+            if (jsMethod == null) {
+                // Not sure if it is possible.
+                invocationExpressionArguments = new[] {thisReference, targetMethod};
+            }
+            else {
+                var jsMethodAccess = targetMethod as JSMethodAccess;
+                var arguments = jsMethod == null
+                    ? null
+                    : jsMethod.Reference.Parameters.Select(
+                            (parameter, index) => (JSExpression) new JSRawOutputIdentifier(parameter.ParameterType, "arguments[{0}]", index))
+                        .ToArray();
+
+                bool isFromDelegate = jsMethod.Reference.Name == "Invoke" && TypeUtil.IsDelegateType(jsMethod.Reference.DeclaringType);
+
+                JSExpression methodInvocation;
+                if (isFromDelegate) {
+                    methodInvocation = targetMethod;
+                }
+                else if (jsMethod.Method.IsStatic) {
+                    methodInvocation = new JSDeferredExpression(JSInvocationExpression.InvokeStatic(jsMethod.Reference.DeclaringType, jsMethod, arguments));
+                }
+                else if (jsMethodAccess == null || jsMethodAccess.IsVirtual) {
+                    methodInvocation = new JSDeferredExpression(JSInvocationExpression.InvokeMethod(jsMethod.Reference.DeclaringType, jsMethod, thisReference, arguments));
+                }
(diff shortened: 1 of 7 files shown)


kg:
```Function.bind``` can't ever be used, for performance reasons. It's real bad. Global deoptimization town. =[

**Comment by iskiselev** [comment]
It should hit only delegate creation. Can you suggest how we can change it?

**Comment by kg** [comment]
If you can, take a look at https://github.com/sq/JSIL/commit/b7df487286d1b5bdf26a597f953534b40371ecf4 , since I solved a similar problem there (and got a huge performance win).

Hand-written closures and CreateNamedFunction are two reasonable choices. They don't seem like they should be faster than bind, but they are. (Though CreateNamedFunction is only faster if you cache the creation function)

**Comment by iskiselev** [comment]
Right now I don't understand how it can help me, but I'll look on it more a little bit later.
In this PR I changed delegate creation from (example):
```
$T02().New(
  $thisType, 
  $thisType.A,
  function () { return JSIL.GetMethodInfo($thisType, "A", $S00(), true); })
```
to:
```
$T02().New(
  $thisType,
  function () { return $thisType.A(); }.bind(this), 
  function () { return JSIL.GetMethodInfo($thisType, "A", $S00(), true); }.bind(this))
```
So, JSIL writes full method invocation inside function (and will use CallVirtual/CallStatic if overload resolution involved).

I suppose it also should fix resolving of MethodInfo in some cases where Generic Parameters are involved (not sure, I've not seen any bug here, but they may presented).

What exactly have performance problem - calling `bind` or calling created function after it?

**Comment by kg** [comment]
The return value of bind is a special class of function. Any time you call a function of that class, the call site is permanently deoptimized for all future calls. (This is a generalization)

**Comment by iskiselev** [comment]
Sorry, still don't understand. Which function is deoptimized:
- function on which bind was called (1)
- function which called bind (2)
- function which call result of bind (3)

(1) is really small call - it will have only function invocation (possible through signature), but called every time we call delegate
(2) in our case is function that created delegate, deoptimization here may be critical
(3) here it is `function Delegate_Invoke () {return method.apply(object, arguments); };`.

I suppose that in (1) and (3) case deoptimization should not be critical, am I wrong?

**Comment by kg** [comment]
I believe both 2 and 3 can be deoptimized. I'm not entirely certain, though. If delegate_Invoke gets deopted that's probably okay, but it could turn into an optimization boundary and prevent inlining, which can be nasty.

**Comment by iskiselev** [comment]
I don't suppose that (2), but we can change it to:
```
$T02().New(
  $thisType,
  (function(thisObject) { function () { return $thisType.A(); }.bind(thisObject)};)(this), 
  (function(thisObject) { function () { return JSIL.GetMethodInfo($thisType, "A", $S00(), true); }.bind(thisObject)};)(this))
```
It will protect (2), but I'm really unsure if we should do it.

**Comment by iskiselev** [comment]
Maybe we should write some test to compare. I'm only unsure how complex should be (2) so that we see any difference from optimization.

File: JSIL/SpecialIdentifiers.cs [hunk 1]
```diff
@@ -111,18 +111,43 @@ JSSpecialIdentifiers js
...
                     new JSType(delegateType),
-                    new JSFakeMethod("New", delegateType, new[] { TypeSystem.Object, TypeSystem.Object }, MethodTypes)
-                ), 
-                method == null ? new [] { thisReference, targetMethod } : new [] {
-                    thisReference, 
-                    targetMethod, 
-                    new JSDefferedExpression(new JSMethodOfExpression(method.Reference, method.Method, method.MethodTypes, method.GenericArguments)),  },
+                    new JSFakeMethod("New", delegateType, new[] {TypeSystem.Object, TypeSystem.Object}, MethodTypes)
+                    ),
+                jsMethod == null
```
kg: Can you split this out into some named locals? This is way too complicated for a single statement.

iskiselev: Done

iskiselev:
Should I write test of performance hit on bind method in this case or @kg, you will check it yourself? Maybe you have better idea about fixing target issue?

kg:
I'm probably going to revisit this at some point and figure out whether it needs to be changed.

iskiselev:
Rebased to current master.

kg:
I'll merge this and we can sort out bind later, I guess.

iskiselev:
Should we create separate open issue to look on it?

iskiselev:
I've broken #880 with this PR.
Problem is that JSDeferredExpression capture all variables of parent function.
Really we need walk through InnerExpression, get all referenced variables and pass them inside some additional function that will generate inner function.
We also can try to change all inner expressions so that they use self instead this if we'll go this way.

iskiselev:
I have prototype of proper fix for both this and #875.
Will prepare PR in few days.

Issue #808, opened 2015-08-21
Roslyn Int64 comparion with 0.


iskiselev:
Roslyn-compiled (release) comparison of Int64 with zero works incorrect - it compares with 0 using ==.
Test case:
```
//@useroslyn
//@compileroption /optimize
using System;

public static class Program {
    public static void Main (string[] args) {
        if (GetZerro() != 0)
        {
            Console.WriteLine("0l != 0");
        }
    }

    public static long GetZerro()
    {
        return 0l;
    }
}
```

PR #809, opened 2015-08-21
Roslyn Int64 comparison

iskiselev:
Fix for #808. Added roslyn test case support using Microsoft.Net.Compilers nuget package and `//@useroslyn` metacomment.


iskiselev:
Looks like I've broken something here. Will look on it more tomorrow.

iskiselev:
Fixed and ready for merge.

iskiselev:
On mono really roslyn meta comment will not have any effect - test still will be compiled with mcs.

iskiselev:
Added fix for #810 (JSIL tried set value with BinOp (=) on JSReadThroughReferenceExpression).
Added some mono compatibility fixes - we are able run translated test cases on Mono now, will do it a little bit later.

iskiselev:
Enabled Translated tests on Travis-CI (all that works on Mono now).
Run tests on Travis in parallel - so build and tests takes about 30 minutes.

Issue #799, opened 2015-08-14
Type.GetTypeInfo() not implemented


dmirmilshteyn:
In particular,
```
The external method 'System.Reflection.TypeInfo GetTypeInfo(System.Type)' of type 'System.Reflection.IntrospectionExtensions' has not been implemented.
```

This makes it particularly annoying as PCL's don't expose extended type information (such as the assembly, declared members, attributes, etc) as part of the standard Type class, and those properties can only be accessed from an instance of TypeInfo. I'm not too sure how to "correctly" implement this, given that it'll likely be deep in the type system used by JSIL... help would be appreciated!

iskiselev:
As workaround for now, your could try select PCL profile that includes Silverlight (if it has enough possibilities for you). All PCL profiles that include Silverlight have full Type class.

iskiselev:
If you want, you could try to implement it yourself in JS files. You need define TypeInfo class and pass Type inside - so that all TypeInfo members will be just proxies to Type members.
Your could also define all TypeInfo/IntrospectionExtensions members in Proxy assembly instead of in JS libraries (probably it is even better option).

dmirmilshteyn:
I'll give it a shot.

To clarify, this is all I would need to do on the proxy side to generate the definition stubs? After that I can just implement all of the methods under ImplementExternals in the JS files? Or should I define all methods and properties in the proxy, and mark them as JSExternal, similar to the TimeSpan proxy?
```csharp
[JSProxy("System.Reflection.TypeInfo", JSProxyMemberPolicy.ReplaceNone, JSProxyAttributePolicy.ReplaceDeclared, JSProxyInterfacePolicy.ReplaceNone, false)]
    [JSStubOnly]
    public class System_TypeInfo
    {
    }
```

iskiselev:
Do you use stubbed or translated BCL?
I suppose that you use stubbed (if you haven't specially created your own .jsilconfig with such settings - you use stubbed). There really will be a problem - as TypeInfo is inside mscorlib and mscorlib is stubbed, even if you provide correct implementation in proxy, it still won't be translated. So, in this case you may still write proxy in Proxies.BCL assembly, than run any test case from \SimpleTestCasesForTranslatedBcl class, then open Tests\Bootstappers\BclBootstrapTranslated.cs.out and copy implementations of TypeInfo/IntrospectionExtensions  to some *.js file that you use.

[...]

PR #813, opened 2015-08-22
Misc. implementations

dmirmilshteyn:
Here is a batch of somewhat-random implementations for missing stuff that I've encountered, as well as a basic verison of TypeInfo (#799)

### DIFF
--- a/JSIL.Libraries/Includes/Core/Reflection/Classes/System.Reflection.Assembly.js
+++ b/JSIL.Libraries/Includes/Core/Reflection/Classes/System.Reflection.Assembly.js
@@ -122,9 +122,17 @@
   );
 
   $.Method({ Static: false, Public: true }, "get_DefinedTypes",
-    (new JSIL.MethodSignature($jsilcore.TypeRef("System.IEnumerable", [$jsilcore.TypeRef("System.Type")]), [], [])),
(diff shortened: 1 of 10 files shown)


kg:
Looks good. If tests pass I'll merge it. Thanks!

dmirmilshteyn:
Guess I didn't test thoroughly enough for the DefinedTypes/IsAssignableFrom implementation. JSIL.GetTypesFromAssembly returns types that are not initialized (__TypeInitialized__ is false). I'm guessing that type initialization is what sets the value of __AssignableTypes__, which is null if used directly from Assembly.DefinedTypes and breaks IsAssignableFrom. Should I check if they have been initialized, and if they haven't, call JSIL.InitializeType on them?

kg:
A type should automatically get initialized before you have access to it. It's fine to call InitializeType on a single type object if you need it to be initialized (like to have that AssignableTypes table). Just don't batch call it on a bunch of types.

For the record, InitializeType runs the static constructor(s).

dmirmilshteyn:
Alright, that should be it!

kg:
This has a bunch of conflicts, probably because I waited too long to merge it. ;-(

dmirmilshteyn:
I'll make a new PR. Merging this with the newly split files will be tricky.

PR #814, opened 2015-08-23
Build improvements

iskiselev:
1. Created NoXNA platform in JSIL.sln (with disabled build of XNA-dependent projects).
2. JSIL_NoXNA.sln and JSIL_NoXNA_NoLibraries.sln deleted - users may use JSIL.sln with NoXNA platform instead.
3. Grunt changed to Gulp.
4. Error in building JSIL.Libraries now will fail build with readable error in log.
5. Removed NuGet dependencies to MSBuild.NodeJs, MSBuild.Npm, MSBuild.Grunt, MSBuild.Gulp and targets from them was copied to JSIL.Libraries\Targets
6. MSBuild/NodeJs integration targets now compatible with Unix build. For now user on Unix still need installed NodeJs/Npm and store path to nodejs in NODEJS environment variable.
7. Clean up of CI integration script - no separate NodeJs task processing anymore, xbuild used instead.


iskiselev:
8. Don't copy NUnit.VisualStudio.TestAdapter and dependecies on Unix, as they breaks test runner on Mono. CI script clean up.

iskiselev:
Added small unrelated to this PR change: provide correct link for failed Translated/Stubbed test, so add EnvironmentVariables/AdditionalFilesToLoad information in link.

iskiselev:
Main improvement here is (3): Error in building JSIL.Libraries now will fail build with readable error in log.
This is why I worked on it, as before errors in processing JSIL.Libraries was just muted.

Issue #812, opened 2015-08-22
BaseType for numeric types


iskiselev:
Currently BaseType for numeric types returns JS String/Number/Boolean, but should return System.ValueType.
I've already fixed it, will do PR later.
This issue is a little bit connected with #427.

Issue #754, opened 2015-07-17
Object check for raw types.


iskiselev:
Is Object check for raw js types (int, string) used by JSIL returns false. Here is test case:
```
using System;
using System.Collections.Generic;
using System.Dynamic;
using System.Linq;

public static class Program
{
    public static void Main(string[] args)
    {
        IsType<object>("qwerty");
        IsType<object>(1);
        IsType<object>(new object());
    }

    public static void IsType<T>(object obj)
    {
        Console.Write(obj is T ? "true" : "false");
    }
}
```

iskiselev:
Really we should do the same for boolean. Will do PR later.

iskiselev:
Really it is duplicate of  #413, but there is more common discussion.

PR #816, opened 2015-08-24
Numeric fixes

iskiselev:
Fixed object check for boolean (#754) and base type for numeric types (#812).

### DIFF
--- a/JSIL.Libraries/Sources/JSIL.Core.Types.js
+++ b/JSIL.Libraries/Sources/JSIL.Core.Types.js
@@ -16,7 +16,7 @@ JSIL.ImplementExternals("System.Object", function ($) {
   $.RawMethod(true, "CheckType",
     function (value) {
       var type = typeof (value); 
-      return (type=== "object" || type === "number" || type === "string");
+      return (type=== "object" || type === "number" || type === "string" || type === "boolean");
     }
   );
 
--- a/JSIL.Libraries/Sources/JSIL.Core.js
+++ b/JSIL.Libraries/Sources/JSIL.Core.js
@@ -302,8 +302,12 @@ JSIL.$GetSpecialType = function (name) {
   JSIL.TypeObjectPrototype.get_Assembly = function() { 
     return this.__Context__.__Assembly__; 
   };
-  JSIL.TypeObjectPrototype.get_BaseType = function() { 
-    return this.__BaseType__; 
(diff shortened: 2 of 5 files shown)


JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -302,8 +302,12 @@ JSIL.$GetSpecialType = function (name) {
...
     return this.__Context__.__Assembly__; 
   };
-  JSIL.TypeObjectPrototype.get_BaseType = function() { 
-    return this.__BaseType__; 
+  JSIL.TypeObjectPrototype.get_BaseType = function () {
+    if (typeof (this.__BaseType__) !== "function")
+      return this.__BaseType__;
+
+    // Workaround for numeric types
+    return $jsilcore.System.ValueType.__Type__;
```
iskiselev: As I know, BaseType could be function (instead of correct Type object) only for numeric types. Probably here should be some more specific check.

JSIL.Libraries/Sources/JSIL.Core.Types.js
```diff
@@ -16,7 +16,7 @@ JSIL.ImplementExternals("System.Object", function ($) {
   $.RawMethod(true, "CheckType",
     function (value) {
       var type = typeof (value); 
-      return (type=== "object" || type === "number" || type === "string");
```
kg: Maybe we should just make this 'return true' :-)

JSIL.Libraries/Sources/JSIL.Core.js
kg: I'm inclined to say that we should never use JS types as basetypes anymore.

JSIL.Libraries/Sources/JSIL.Core.Types.js
iskiselev: Probably we should not return true for `undefined`. Also we should implement Object methods for raw JS object for it.

kg:
Crash after merging on travis... weird

https://travis-ci.org/sq/JSIL/jobs/77078965

iskiselev:
It crashes sometimes. Just run once again if it has crashed :(

kg:
Yeah, just got an assert in mono sgen. Maybe we should file a bug...

iskiselev:
Really I've no ideas how to get some additional info for filling bug report. Pretty sure, that native stack trace will be not enough.

PR #819, opened 2015-08-25
Reflection additions

iskiselev:
1. Added System.Reflection.TypeInfo, System.RuntimeType derived from it now.
2. typeof(T).GetType() returns typeof(System.RuntimeType)
3. JSIL.GetMemberAttributes inherit support for MethodInfo.
4. GetConstructors always use DeclaredOnly internally (test added).
5. MemberInfo.IsDefined impelemented.
6. PropertyInfo GetValue/SetValue.
7. Type.IsClass implemented.

### DIFF
--- a/JSIL.Libraries/Includes/Core/Reflection/Classes/System.Type.js
+++ b/JSIL.Libraries/Includes/Core/Reflection/Classes/System.Type.js
@@ -464,13 +464,6 @@
       }
     );
 
-    $.Method({ Public: true, Static: false }, "GetType",
-      new JSIL.MethodSignature($.Type, []),
-      function () {
-        return $jsilcore.System.Type.__Type__;
-      }
-    );
-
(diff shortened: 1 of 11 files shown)


JSIL.Libraries/Includes/Core/Reflection/Classes/System.Type.js
```diff
@@ -328,7 +335,7 @@
           $jsilcore.TypeRef("System.Array", [$jsilcore.TypeRef("System.Reflection.ParameterModifier")])
       ], []),
       function GetConstructor(bindingAttr, binder, callConvention, types, modifiers) {
-        return getConstructorImpl(this, bindingAttr, types);
+        return getConstructorImpl(this, bindingAttr | System.Reflection.BindingFlags.DeclaredOnly, types);
```
kg: The bitwise or operator on enums in JSIL doesn't necessarily do what you want, since it produces an int. This is probably fine, though...

PR #829, opened 2015-08-27
Translated scripts should contain BOM at the beginning.

iskiselev:
Small fix to preserve BOM.


kg:
I'm actually getting failures running the generated scripts in Chrome because it's choking on the BOM. It looks like the problem is that the BOM comes after the comment header.

Does this PR fix that?

kg:
Looks like it does. Thanks.

Issue #76, opened 2012-03-20
support for SourceMaps


clarity99:
It seems Chrome has added support for SourceMaps (https://plus.google.com/u/1/110412141990454266397/posts/Nvr6Se6eAPh, http://code.google.com/p/google-web-toolkit/wiki/SourceMaps) and Firefox seems to be working on it as well (https://wiki.mozilla.org/DevTools/Features/SourceMap). 

I think support for this would be a great addition to JSIL, as knowing the original source code location  would simplify debugging a lot.

iskiselev:
Do anybody knows some tool that allow you dynamically switch between SourceMap provided view and original JavaScript right inside debug session? So, that it preserve current debug context - current line, stack, breakpoints?
Otherwise source C# view will be not usable for a lot of tasks (but could help a little bit).

iskiselev:
SourceMap support could be activated by adding `"BuildSourceMap" : true` in .jsilconfig.

iskiselev:
Probably we need introduce something like debug translating options, that will skip clever optimizations (such as EliminateTemporaries, probably some others). Also we may need preserve generation of nop as ";" lines - it will help build more precious source map.
@kg - do you have ideas which FunctionTransformPipeline steps are not necessary and was introduced only for speed improvements? We may try to prepare list of them here (or in wiki).

kg:
At this point I think the vast majority of them are performance-related. It is possible some behaviorally-important passes depend on the output of optimization passes, though.

I think a debug translation mode is a good idea.

PR #852, opened 2015-09-14
Source map initial support

iskiselev:
Looks like I was able add some initial support of Source maps (#76). For now it was done not very clean and mostly hackish, but it already works in Chrome/IE/Visual Studio.
To enable their generation, user need add `BuildSourceMap : true` key in .jsilconfig.

### DIFF
--- a/JSIL/AST/JSAstVisitor.cs
+++ b/JSIL/AST/JSAstVisitor.cs
@@ -19,21 +19,6 @@ namespace JSIL.Ast {
             public string ChildName;
         }
 
-        public class NodeProcessedEventArg : EventArgs
-        {
-            public JSNode Node { get; private set; }
-
-            public NodeProcessedEventArg(JSNode node)
-            {
-                if (node == null)
-                {
-                    throw new ArgumentNullException("node");
-                }
-
-                Node = node;
-            }
-        }
-
         public const int DefaultStackSize = 32;
 
         public readonly Stack<JSNode> Stack = new Stack<JSNode>(DefaultStackSize);
@@ -55,7 +40,7 @@ namespace JSIL.Ast {
 
         protected bool VisitNestedFunctions = false;
 
-        protected event EventHandler<NodeProcessedEventArg> AfterNodeProcessed;
+        protected event Action<JSNode> AfterNodeProcessed;
 
         protected JSAstVisitor () {
             Visitors = VisitorCache.Get(this);
@@ -236,7 +221,7 @@ namespace JSIL.Ast {
 
             if (node != null && AfterNodeProcessed != null)
             {
-                AfterNodeProcessed(this, new NodeProcessedEventArg(node));
+                AfterNodeProcessed(node);
             }
         }
 
--- a/JSIL/AST/JSAstVisitor.cs
+++ b/JSIL/AST/JSAstVisitor.cs
@@ -19,6 +19,21 @@ namespace JSIL.Ast {
             public string ChildName;
         }
 
+        public class NodeProcessedEventArg : EventArgs
+        {
+            public JSNode Node { get; }
+
+            public NodeProcessedEventArg(JSNode node)
+            {
+                if (node == null)
+                {
+                    throw new ArgumentNullException("node");
+                }
+
+                Node = node;
+            }
+        }
+
         public const int DefaultStackSize = 32;
 
         public readonly Stack<JSNode> Stack = new Stack<JSNode>(DefaultStackSize);
@@ -39,6 +54,8 @@ namespace JSIL.Ast {
         protected readonly VisitorCache Visitors;
 
         protected bool VisitNestedFunctions = false;
+
+        protected event EventHandler<NodeProcessedEventArg> AfterNodeProcessed;
 
         protected JSAstVisitor () {
             Visitors = VisitorCache.Get(this);
@@ -169,7 +186,7 @@ namespace JSIL.Ast {
         /// </summary>
         /// <param name="node">The node to visit.</param>
         /// <param name="name">The name to annotate the node with, if any.</param>
-        public virtual void Visit (JSNode node, string name = null) {
+        public void Visit (JSNode node, string name = null) {
             if (node is JSFunctionExpression) {
                 // HACK: No better place to put this at present.
                 // AST visitors shouldn't recurse into nested functions because those function(s)
@@ -215,6 +232,11 @@ namespace JSIL.Ast {
 
                 NodeIndex = oldNodeIndex;
                 StatementIndex = oldStatementIndex;
+            }
+
+            if (node != null && AfterNodeProcessed != null)
+            {
+                AfterNodeProcessed(this, new NodeProcessedEventArg(node));
             }
         }
 
--- a/JSIL/AST/JSAstVisitor.cs
+++ b/JSIL/AST/JSAstVisitor.cs
@@ -21,7 +21,7 @@ namespace JSIL.Ast {
 
         public class NodeProcessedEventArg : EventArgs
         {
-            public JSNode Node { get; }
+            public JSNode Node { get; private set; }
 
             public NodeProcessedEventArg(JSNode node)
             {
--- a/JSIL/AST/JSAstVisitor.cs
+++ b/JSIL/AST/JSAstVisitor.cs
@@ -169,7 +169,7 @@ namespace JSIL.Ast {
         /// </summary>
         /// <param name="node">The node to visit.</param>
         /// <param name="name">The name to annotate the node with, if any.</param>
-        public void Visit (JSNode node, string name = null) {
+        public virtual void Visit (JSNode node, string name = null) {
             if (node is JSFunctionExpression) {
                 // HACK: No better place to put this at present.
                 // AST visitors shouldn't recurse into nested functions because those function(s)
--- a/JSIL/ILBlockTranslator.cs
+++ b/JSIL/ILBlockTranslator.cs
@@ -13,6 +13,7 @@ using JSIL.Internal;
 using JSIL.Transforms;
 using Microsoft.CSharp.RuntimeBinder;
 using Mono.Cecil;
+using Mono.Cecil.Cil;
 
 using TypeInfo = JSIL.Internal.TypeInfo;
 
@@ -22,6 +23,7 @@ namespace JSIL {
         public readonly DecompilerContext Context;
         public readonly MethodReference ThisMethodReference;
         public readonly MethodDefinition ThisMethod;
+        public readonly MethodSymbols Symbols;
         public readonly ILBlock Block;
         public readonly JavascriptFormatter Output = null;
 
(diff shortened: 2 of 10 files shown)


JSIL/AST/JSAstVisitor.cs
```diff
@@ -169,7 +169,7 @@ protected class Adapter<TVisitor, TNode>
         /// </summary>
         /// <param name="node">The node to visit.</param>
         /// <param name="name">The name to annotate the node with, if any.</param>
-        public void Visit (JSNode node, string name = null) {
+        public virtual void Visit (JSNode node, string name = null) {
```
kg: Making ```Visit``` virtual is nasty, this method gets called a *lot* by every single transform.

Since you're just using this to run a hook at the end of ```Visit``` for every node in ```AstEmitter```, could you move to a default-null delegate (or event) on ```AstVisitor``` that is called if not-null at the end of ```Visit```? That should be near-zero-cost for the other visitors.

JSIL/SourceMapBuilder.cs
```diff
@@ -0,0 +1,277 @@
...
+                return Equals(left, right);
+            }
+
+            public static bool operator !=(Mapping left, Mapping right)
+            {
+                return !Equals(left, right);
+            }
+        }
+
+        private static class Base64Vlq
```
kg: Is this someone else's code? If so, is it a compatible license? Just checking.

kg:
Cool! The last time I checked this didn't seem possible because ILSpy was dropping ILRanges in a bunch of places. I'm impressed you got this working and did so with such a small amount of code :-)

JSIL/ILBlockTranslator.cs
```diff
@@ -1130,11 +1138,28 @@ bool forDynamic
...
             var statement = translated as JSStatement;
-            if (statement == null) {
-                var expression = (JSExpression)translated;
+            if (statement == null)
+            {
+                var expression = (JSExpression) translated;
 
                 if (expression != null)
+                {
+                    var ranges =
```
kg: This pair of selects looks like it will perform really badly, so I'd like the source map support to be behind a configuration bool.

JSIL/ILBlockTranslator.cs
```diff
@@ -1130,11 +1138,28 @@ bool forDynamic
...
+                        node.GetSelfAndChildrenRecursive<ILExpression>(item => true).SelectMany(item => item.ILRanges);
+                    var symbolInfo = Symbols != null
+                        ? ranges
+                            .SelectMany(
+                                range =>
+                                    Symbols.Instructions.Where(
+                                        item => range.From <= item.Offset && item.Offset < range.To))
+                            .Select(item => item.SequencePoint)
+                            .Where(item => !(item.StartLine == item.EndLine && item.StartColumn == item.EndColumn))
+                            .ToArray()
```
kg: Use ```.ToList```, it's much faster (counter-intuitively).

JSIL/ILBlockTranslator.cs
```diff
@@ -1130,11 +1138,28 @@ bool forDynamic
...
-            if (statement == null) {
-                var expression = (JSExpression)translated;
+            if (statement == null)
+            {
+                var expression = (JSExpression) translated;
 
                 if (expression != null)
+                {
+                    var ranges =
+                        node.GetSelfAndChildrenRecursive<ILExpression>(item => true).SelectMany(item => item.ILRanges);
```
kg: If you could do this without ```GetSelfAndChildrenRecursive``` that would be good. I forget if they expose a way to treewalk directly. ```GetSelfAndChildrenRecursive``` is one of the biggest timewasters in JSIL profiles and it allocates a lot; in this case we're just going to be throwing its result away to begin with.

JSIL/SourceMapBuilder.cs
iskiselev: Yes, it's Mozilla code (https://github.com/mozilla/source-map/), to be license compliant we need add header from: https://github.com/mozilla/source-map/blob/master/lib/base64-vlq.js.

JSIL/ILBlockTranslator.cs
iskiselev: Sure, I'm even added configuration for it - just need do some more cleanups.

iskiselev:
Main problem that I see now - it that IL in pdb annotated only statement-based and not IL-command level. It may be good idea write some Roslyn extension, that will preserve information for each IL-command.

JSIL/ILBlockTranslator.cs
iskiselev: Probably it would be better always preserve source ILRange in  JSNode on this stage, and introduce separate pass over converted AST representation to build SourceMap if needed, as this processing could be really more complicated.
Right now JSIL may reuse JSNode (as variables) - and it may be a problem for this task. Will investigate a little bit more proper point of ILTranslation extension.
Will be good if right solution would not require us rewrite each JSNode creation.

JSIL/JavascriptAstEmitter.cs
```diff
@@ -78,6 +80,16 @@ Configuration configuration
             this.Visit(node);
         }
 
+        public override void Visit(JSNode node, string name = null)
+        {
+            base.Visit(node, name);
+            if (Output.SourceMapBuilder != null && node != null && node.SymbolInfo != null && node.SymbolInfo.Any())
+            {
+                Output.SourceMapBuilder.AddInfo(Output.Output.Line, Output.Output.FirstNonSpace, node.SymbolInfo);
```
iskiselev: Really here should be `Output.Output.Column` instead of `Output.Output.FirstNonSpace`. But VS debugger doesn't work if there is no symbol for first non-space symbol on line.
While we still has only one point per line (while we take info from PDB), we could use this hack. We need some better solution in general.

JSIL/ILBlockTranslator.cs
iskiselev: Will look on it later - really this point will change a lot while Source map support will be more mature.

iskiselev:
Updated based on PR input.

JSIL/AST/JSAstVisitor.cs
```diff
@@ -19,6 +19,21 @@ public struct EnclosingNode<T> {
             public string ChildName;
         }
 
+        public class NodeProcessedEventArg : EventArgs
```
kg: Can you make this a struct or just pass a JSNode directly? 'sender' is useless here so there's not much point in paying for a heap allocation per-node.

JSIL/AST/JSAstVisitor.cs
```diff
@@ -216,6 +233,11 @@ protected class Adapter<TVisitor, TNode>
                 NodeIndex = oldNodeIndex;
                 StatementIndex = oldStatementIndex;
             }
+
+            if (node != null && AfterNodeProcessed != null)
```
kg: Huh, node can be null?

kg:
I'm not sure I understand the TranslationResult changes. Why does it need a new argument to AddFile? Is there a reason I'm overlooking why source maps can't just be additional files added to the TranslationResult? I don't like the idea of adding additional logic to TranslationResult, especially logic to generate new content on-demand lazily.

JSIL/AST/JSAstVisitor.cs
iskiselev: I've seen it when I translate my application. I've a lot of log messages, but as it doesn't affect anything I've not looked on it.

kg: OK. Probably indicates a bug on my end. Don't mind having the conditional there, though.

JSIL/AST/JSAstVisitor.cs
iskiselev: Here is common question - should we use standard event pattern (sender, eventArgs) - or could we optimize it.
In most scenarios I prefer use standard pattern as it simplify code readability for other developers.
Here, if you'd like it - let's change event signature to just `Action<JSNode>`

iskiselev:
Source map generation is relative to generated javascript.
If we use some post-processing of TranslationResult (merge it, as an example) - we should also merge  SourceMap and generate it only afterwards.
Right now I've not yet supported SourceMap merging, but we may need it.

JSIL/AST/JSAstVisitor.cs
kg: Normally sticking with convention is good, but an allocation per node visited is pretty nasty perf-wise. GC is a major bottleneck in JSIL :(

kg:
I see. I'd prefer that we not put this in TranslationResult directly, and put this logic somewhere else. It's intended that TranslationResult just be a dumb container for already-translated files, so the way it does .Build on-demand is problematic.

iskiselev:
Could you suggest home for this intermediate SourceMap in this case. So it relates to JavaScript stream that contains in ResultFile and make no sense without it - but should be changed if we do any post-processing for ResultFile (add header, merge and so on)?

kg:
My instinct is that we should have a type representing all the source mappings for the current compile, and then at the end when we write out TranslationResult there's a way to write out the source maps.

The source maps don't go into the compile manifest (JSIL.js has no reason to load them), etc. And we may want to just produce one source map, and we might not want to output it to disk.

Your point about needing to sync it up with the JS in ResultFile is a good one though...

iskiselev:
According to sourcemap standard we can't do one sourcemap for several .js files - so it always be one/zero map per generated *.js.
Moreover, we need modify generated *.js to attach sourcemap to it - we add:
`//# sourceMappingURL=file://...`
to the end of each JS if we've generated sourcemap.

kg:
Can we programmatically decide the name of the sourcemap up front, and just add it to the file header when sourcemaps are enabled? Then there's no need to filter the results or do dynamic code generation in TranslationResult.

iskiselev:
Yes we can. Let's summarize all restrictions here:
1. There could be only one sourcemap per *.js file (not sure, probably I'm not correct here).
2. One sourcemap provide info for only one *.js file
3. Source map and js knows each other names.
4. If js file was edited (especially lines added/removed), sourcemap needs to be rebuild.

iskiselev:
But I still don't understand how we will solve problem with merging TranslationResult if we generate sourcemap upfront.

kg:
Sorry, I don't think we need to generate the map up front, just assign it a name up front. I'm ok with the sourcemap being generated on demand, I just don't want the on-demand generation to be a part of translationresult. does that make sense?

iskiselev:
Unfortunately here is still a problem.
We may merge TranslationResult. If we generate SourceMap reference upfront, we will broke restriction (1).
So, looks like it is responsibility of component that build SourceMap to add reference back to original file.

kg:
I'll think more about how we should do this. Can you explain the merging thing to me? I'm not aware of JSILc ever merging files.

iskiselev:
JSIL merge all TranslationResult when generate test cases. On other hand probably we may ignore it.

kg:
Ahh, right. The TranslationResult merging is Bad and it probably shouldn't do it. We can ignore that when designing new features; if it becomes a problem I'll just fix it.

iskiselev:
@kg, after my test, it looks that best option to reference SourceMap is using URL with full local path (through FILE://).
It is only option that will work on developer computer with any web server configuration - and in most cases developer needs them on his local computer.

But there is a problem - currently AssemblyTranslator doesn't know where scripts will go - it works with translation results. How could you advice me to solve this problem?

iskiselev:
Still leave writing SourceMap link on Translation result writing to file, but cleaned it a little bit.

kg:
Egh. Writing file:// into the source to point to the sourcemap is really gross. I'm not sure there's a sensible way to make that work... I'll think about it.

kg:
Can you look into this?
https://ci.appveyor.com/project/kg/jsil/build/1.0.64

iskiselev:
Error looks really strange. Can you restart build to be sure if it is constantly reproducible?

Issue #855, opened 2015-09-17
Create UInt64 from UInt32


iskiselev:
JSIL translated code fails on next test case:
```
using System;

public static class Program {
    public static void Main (string[] args)
    {
        ulong l = 0x80000000;
        Console.WriteLine(l);
    }
}
```

It fails with `Error: cannot construct UInt64 from negative number`.
This code was translated to:
```
var $T02 = function () {
    return ($T02 = JSIL.Memoize($asm01.System.UInt64)) ();
  };
    $T01().WriteLine($T02().FromNumber(-2147483648));
```

So, probably we should allow creation of UInt64 from negative numbers with built-in conversion or fix how JSIL translates it.

markusjohnsson:
Yup, FromNumber should probably interpret -2147483648 as a 32 bit int, take the bits from that and treat it as a 32 bit uint. 

See conv_u8: https://github.com/markusjohnsson/cil.js/blob/master/CilJs/Runtime/Runtime.js#L363

Issue #850, opened 2015-09-11
JSReplacement and reflection


iskiselev:
Currently using of JSReplacement  affect reflection, as member with JSReplacement is not translated.
I suggest in any translate JSReplacement  method with same body as passed in JSReplacement attribute.

kg:
The replacement can't be a method body, though, because it's a textual replacement macro.

Not sure what we should do about this...

iskiselev:
Really, JSReplacement currently do two things:
1. Provide macro to use instead of method.
2. Suppress marked method from output.

I'd like split it a little bit. JSReplacement still will replace method usage, but will not suppress marked method generation (but it still not be called in most cases, except reflection).
Suppress marked method from output should do some other meta attribute or some flag inside JSReplacement.

kg:
I'm OK with making JSReplacement not strip the method from output, as long as it doesn't break anything. Seems like a good idea.

The behavioral gap between normal invocation/reflection invocation will suck, though.

iskiselev:
That's why I originally suggested put same macro inside MethodBody that was marked with JSReplacement. So, body of method marked with JSReplacement included in translated results will be just same if you use it through reflection.

kg:
Well, there are cases where it wouldn't work, but maybe that's an acceptable compromise.

iskiselev:
On other side, right now we use JSReplace mostly to change invocation format if we cannot use standard - such as calling string/array/numeric methods.
Here, really, you may even prefer preserve original method body and just use some hacks to call it.
So, it really good candidate to be spitted into two separate attributes.

kg:
I think you're right. We want a split where one of them is 'replace method body (and inline optimistically)' and the other one is a macro replacement of sorts.

iskiselev:
@kg, what do you think about JSSuppressOutput meta attribute?

Right now we have:
JSIgnore - will affect both method output and method invocation
JSExternal - affect only method output, but don't allow fully suppress method from output, only method body
JSSuppressTypeDeclaration - will write type alias. Probably should be renamed to something more descriptive.

We still have no any attribute, that will affect only method output and will just mute it for some particular class/method/etc.

If JSReplacement will not suppress output, there may be some places where we want mark method additionally with JSSuppressOutput attribute.

iskiselev:
Looks like we should do changes proposed in this issue to implement ToString for char (#938)

iskiselev:
[...]

Issue #507, opened 2014-05-27
Move TypeDeclarationsToSuppress in meta attributes


iskiselev:
Currently JSIL have list of types, which output to .js files will be suppressed. They are declared in AssemblyTranslator.TypeDeclarationsToSuppress. We can't achieve same behavior for any other type using meta attributes.

I think it may be good idea to introduce new meta attribute with same meaning and build this list dynamically.

I want to introduce, because I found small trouble with DCE:
I create manual implementation for one class in js. This class contains private static field. As I have manual implementation for such class, I marked it with JSStubOnly attribute. According to this, DCE doesn't analyze it's members - so it think that private static field of this class never used and strip it from output. As we still write JSIL.MakeType for this type, it overrides my manual description containing this field. If I was able to suppress writing this class fully, it will use my manually created class description with this static field.

Other solution (and maybe it will be good to support also) will be support field declaration inside JSIL.ImplementExternals.

PR #857, opened 2015-09-18
Bootstrap from C#. Number formats

iskiselev:
Per discussion in #850.
Added new C# project - JSIL.mscorlib.
This project is translated on JSIL compilation and goes into Libraries folder. It is attached as bootstrap through JSIL.js.
Currently I've added (Numeric).ToString(format, ...) implementations through mono System.NumberFormatter.

Also this PR contains several StringBuilder method implementations, fix for #855.

To support Bootstrap from C# there was introduced several meta attributes:
JSRepaceAssemblyDeclaration - replaces JSIL.DeclareAssembly with passed string 
JSOverrideAssemblyReference - overrides $asm... for passed type assembly with string and enclose all translation in self-execution function
JSImportType - include type from other assembly in translation as it was defined here. Should be use with JSProxy.
JSSuppressTypeDeclaration (#507) - TypeDeclarationsToSuppress moved here.

@kg, please wait for tests to pass. I've not yet check all of them - but we can already start review of this change.


JSIL.Libraries/Includes/Bootstrap/Resources/Main.js
```diff
@@ -8,7 +8,7 @@ if (!$jsilcore)
 
 $jsilcore.getCurrentUICultureImpl = function () {
   var language;
-  var svc = JSIL.Host.getService("window", false);
+  var svc = JSIL.Host.getService("window", true);
```
kg: Why?

iskiselev: We need initialize current culture even in non-browser environment.
Really next line even tries to do it:
```
  if (svc) {
    language = svc.getNavigatorLanguage() || "en-US";
  } else {
    language = "en-US";
  }
```

But in non-browser environment it will not return null, but fail.
Now it will return null.

kg: Ah, I probably got the second argument's boolean nature backwards... bools are awful.

iskiselev:
Currently JSImport works incorrectly on Linux.
Some methods of imported classes become UntranslatableFunctions. Right now will look on it.
I've now worked on Linux a lot before. @kg, what could you advice to debug JSIL on Linux (except of improving log output)?

kg:
I don't really know, I find Linux development a nightmare. None of the tools are good. That's why some of the 'doesn't work on Mono' tests are still broken.

iskiselev:
@kg, how could function go into FunctionCache?
On linux, when I'm try to translate imported functions, looks like some of them are already in cache with null records.
I start suppose that either JSIL key comparer works incorrectly - or Concurrent Dictionary returns incorrect records.

iskiselev:
It was my fault. Functions went to cache through GetSecondPass.
Last that I still don't understand - why it wasn't happen on windows (both mono and .net)?

kg:
Mono's ConcurrentDictionary is probably broken. They've had trivial bugs in their BCL containers before, and they take forever to get fixed (even when I provide a fix). It's a nightmare.

iskiselev:
Formatting test on travis still failing, but I've just tested it on my own Linux box - and it works as expected.
For now I'll mark this test (IntegerToStringFormatted) as FailsOnMono.

We need really upload somewhere all artifacts (at least Libraries folder and all tests *.out files) to be able test problems that happens on Travis - but it is another task.

iskiselev:
So, now I think this PR is stable enough and could be merged.
Probably we need write some additional checks for JSImport edge cases, but it may be better to do later - when we'll have some more experience with translated Bootstrap.

iskiselev:
Ops. Looks like I missed non-integer types. Let's see test results.

iskiselev:
Get artifacts from travis and looked why test fails there.
Looks like NumberFormatter.Format, that contains switch with goto translated incorrectly if was compiled under mono.

iskiselev:
@kg, could you review/discuss this? I'm working now on new implementation for #773, and it is done on top of it. It will be not so easy to make that PR fully independent, but I don't want make this PR too complicated.

JSIL.mscorlib/JSIL/System/NumberFormatter.Data.cs
```diff
@@ -0,0 +1,916 @@
+﻿//
+// mono/mono/metadata/number-formatter.h
```
kg: Please explicitly call out what license that was under in addition to the filename. Doesn't need the whole license prologue

JSIL/JavascriptFormatter.cs
```diff
@@ -1164,10 +1167,30 @@ select GetNameOfType(ga as dynamic)).ToArray())
                 Space();
                 Identifier(PrivateToken.IDString);
                 WriteRaw(" = ");
-                WriteRaw("JSIL.DeclareAssembly");
-                LPar();
-                Value(Assembly.FullName);
-                RPar();
+                if (AssmeblyDeclarationReplacement == null)
```
kg: Can you fix that typo?

Proxies/BCL/System.cs
```diff
@@ -28,4 +31,31 @@ public class System_Security_Cryptography_X509Certificates_X509Certificate2
...
+    [JSProxy(typeof(DateTimeFormatInfo), JSProxyMemberPolicy.ReplaceDeclared, JSProxyAttributePolicy.ReplaceDeclared, JSProxyInterfacePolicy.ReplaceNone, false)]
+    public class System_Globalization_DateTimeFormatInfo
+    {
+        [JSReplaceConstructor]
+        static System_Globalization_DateTimeFormatInfo()
+        {
+        }
+    }
+
+    /*[JSProxy(typeof(ExecutionContext), JSProxyMemberPolicy.ReplaceNone, JSProxyAttributePolicy.ReplaceDeclared, JSProxyInterfacePolicy.ReplaceNone, false)]
```
kg: Why is this commented out?

kg:
Took a second look at this and I don't see anything wrong other than the two nits I commented on. Thanks.

JSIL.mscorlib/JSIL/System/NumberFormatter.Data.cs
iskiselev: Hmm, good catch. Looks like this file, as it located in mono/mono folder is licensed only with LGPL or commercial license.
It is really bad, so we should check if it was previously located under anther directory pr ask them for using this file with another license :(

Proxies/BCL/System.cs
iskiselev: Looks like I've added it by mistake. Removed.

iskiselev:
PR fix applied. Also added double signature fix and .Parse(string, IFormatProvider) signature as it is used by NumberFormatter.

PR #858, opened 2015-09-19
F# CodeDom updated

iskiselev:
Updated F# CodeDom (they fixed fsprojects/FSharp.Compiler.CodeDom#8, so it works on Linux now).
Install FSharp compiler and NUnit test runner in packages folder (it should be cached in your environment).
Removed FailsOnMono from F# tests and one more fixed test.


iskiselev:
@kg, will you merge it as it is starting point with my next PR (using travis cache and artifacts uploading)?
This PR should be safe. You could rerun failed tests, but really there were some problems not connected with this PR.

kg:
Sorry I forgot about this. Gonna rerun the failed travis build and then merge if it passes. Thanks.

PR #863, opened 2015-10-09
Verbatium.Expression generic

iskiselev:
Introduced generic forms for Verbatium.Expression to prevent boxing.
It will help solve problem that I described in #773.

Issue #860, opened 2015-10-04
Unsafe pointer from string


iskiselev:
Looks like JSIL doesn't support resolving pointer from string. Here is test case:
```
using System;

static class Program {
    public static void Main () {
        string str = "qww";
        unsafe
        {
            fixed (char* sPtr = str)
            {
                *sPtr = 'a';
            }
        }

        Console.WriteLine(str);
    }
}
```

Translation result:
```
    var sPtr = JSIL.UntranslatableInstruction("Implicit conversion to pointer: Untranslatable Expression Conversion of expression '@<var System.String str>' to pointer");
    sPtr.set(("a").charCodeAt(0));
    $T00().WriteLine("qww");
```

iskiselev:
It would be very good, if at least read operations will work.

iskiselev:
Looks like it will be pretty easy to implement for read operations. Will do it.
In any case, mutate string through a pointer is not best idea even in .Net.

kg:
Yeah, PinAndGetPointer just needs to copy the string's chars into a temporary utf-16 buffer.

iskiselev:
Will it be OK if we create Pointer with null as MemoryRange and source string as view (I've already done it)? Or there will be problems if we'd like to cast pointer afterward?

kg:
No, that's no good.

iskiselev:
What should we do on write through such pointer attempts? Right now I throw NotImplementedException. Or should we allow write to such temporary array (waithout writing it back to source string)?

kg:
I would allow the write and not represent it in the source string. It's not possible to mutate a string instance in JS anyway.

While mutating a pinned string in C# might work it's a terrible idea so I'm okay with not implementing it. If it's needed, we could make it work by implementing the unpin portion of fixed blocks and using that to flush the change to the string somehow... but the string's object reference would change, so it's still not gonna work.

iskiselev:
Additional notes: get pointer to char[] element doesn't work also.
One more thing to test - cast pointer to byte* or int* (as sizeof(char)=2).
```
using System;

static class Program {
    public unsafe static void Main(string[] args)
    {
        string str = "qww";
        unsafe
        {
            fixed (char* sPtr = str.ToCharArray())
            {
                RunMe(sPtr, str.Length);
            }
        }

        Console.WriteLine(str);
    }

    private unsafe static void RunMe(char* sPtr, int length)
    {
        char* endPtr = sPtr + length;
[...]

PR #864, opened 2015-10-10
Box for Numeric types.

iskiselev:
New version for numeric type boxing. This PR should be now looked instead of #773.

### DIFF
--- a/JSIL.Libraries/Includes/Reference/Classes/JSIL.Box.js
+++ b/JSIL.Libraries/Includes/Reference/Classes/JSIL.Box.js
@@ -1,0 +1,69 @@
+﻿JSIL.MakeClass("System.Object", "JSIL.Box", true, ["TValue"], function ($) {
+  function prepareInterfaceCaller(interfaceMethod) {
+    return function() { return interfaceMethod.apply(this.value, arguments); };
+  }
+
+  $.SetValue("__IsBox__", true);
+
+  $.Method({ Static: false, Public: true }, ".ctor",
+    (new JSIL.MethodSignature(null, [new JSIL.GenericParameter("TValue", "JSIL.Box")], [])),
+    function _ctor(value) {
+      this.value = value;
+    }
+  );
+
+  $.RawMethod(true, "$addTypeMethods",
+    function $addTypeMethods(ctor, type) {
+      var ctor = ctor.prototype;
+      var type = type.prototype;
+      //var initType = new ctor(null);
+
+      for (var key in type) {
+        if (type.hasOwnProperty(key) && !(key in ctor)) {
+          var obj = type[key];
+          if (typeof (obj) == "function") {
+            ctor[key] = prepareInterfaceCaller(obj);
+          }
+        }
+      }
+    }
+  );
+
+  $.RawMethod(false, "valueOf",
+    function valueOf() {
+      return this.value.valueOf();
+    }
+  );
+
+  $.RawMethod(false, "toString",
+    function toString() {
+      return this.value.toString();
+    }
+  );
+
+  $.RawMethod(false, "GetHashCode",
+    function Box_GetHashCode() {
+      return JSIL.ObjectHashCode(this.value, true, this.TValue);
+    }
+  );
+
+  $.RawMethod(true, "IsBoxedOfType",
+    function isBoxedOfType(value, type) {
+      return value !== null && value !== undefined && value.__IsBox__ && value.TValue.__TypeId__ === type.__TypeId__;
+    }
+  );
+
+  $.Method({ Static: false, Public: true }, "Object.Equals",
+    new JSIL.MethodSignature($.Boolean, [$.Object], []),
+    function Box_Equals(other) {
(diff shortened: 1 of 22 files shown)


Proxies/Numbers.cs
```diff
@@ -29,7 +29,7 @@ public abstract class IntegerProxy {
         }
 
         [JSIsPure]
-        [JSReplacement("($this === $rhs)")]
+        [JSReplacement("(($this).valueOf() === ($rhs).valueOf())")]
```
kg: I think .valueOf() will force the creation of an object wrapper, so you probably want to do ```x | 0 === y | 0``` for signed ints, ```x >>> 0 === y >>> 0``` for unsigned ints, and ```+x === +y``` for floats.

kg: Oh, though one caveat is that ```x >>> 0 === y >>> 0``` will be wrong if one of the operands is signed, since it'll force it to unsigned. Hm.

iskiselev: Really we should split it into to separate.
T.Equals(T) should be left as $this === $rhs
T.Equal(object) should be $this === $rhs.value().
Hm, really there may be null or even undefined. Than I think we don't need override in special way `T.Equal(object)` for numeric type. It will work slow, but it should not be called often.

iskiselev:
The only failing test will be fixed with #863.

iskiselev:
Fixed Equals signature based on discussion. Let's wait test cases (`UnsafeIntPerformanceComparison` failing is OK).

JSIL.Libraries/Includes/Reference/Classes/JSIL.Box.js
```diff
@@ -0,0 +1,69 @@
...
+      var type = type.prototype;
+      //var initType = new ctor(null);
+
+      for (var key in type) {
+        if (type.hasOwnProperty(key) && !(key in ctor)) {
+          var obj = type[key];
+          if (typeof (obj) == "function") {
+            ctor[key] = prepareInterfaceCaller(obj);
+          }
+        }
```
iskiselev: Here is hack we need it to implement original type interfaces.

JSIL/ILBlockTranslator.cs
```diff
@@ -3158,7 +3174,12 @@ protected JSExpression Translate_Unbox(ILExpression node, TypeReference targetTy
 
         protected JSExpression Translate_Box (ILExpression node, TypeReference valueType) {
             var value = TranslateNode(node.Arguments[0]);
-            return JSReferenceExpression.New(value);
+            var originalType = value.GetActualType(TypeSystem);
+            var refenrence = JSReferenceExpression.New(value);
+            // Hack, but I don't know how get System.Decimal reference.
+            return (_rawTypes.Contains(originalType) || originalType.FullName == "System.Decimal")
```
iskiselev: Check for decimal should go away with #871 fix.

JSIL/ILBlockTranslator.cs
```diff
@@ -95,6 +97,20 @@ public class ILBlockTranslator {
...
+                TypeSystem.SByte,
+                TypeSystem.Byte,
+                TypeSystem.Int16,
+                TypeSystem.UInt16,
+                TypeSystem.Int32,
+                TypeSystem.UInt32,
+                TypeSystem.Single,
+                TypeSystem.Double,
+                TypeSystem.Char
+            };
```
iskiselev: I've seen, that int64 could use raw type also based on configuration. We probably need check it here?

JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Decimal.js
```diff
@@ -151,4 +151,6 @@
   $.Field({ Static: false, Public: false }, "value", mscorlib.TypeRef("System.Double"), function () {
     return 0;
   });
+  JSIL.MakeBoxMethod($);
+
```
iskiselev: This should go away with #871 fix.

iskiselev:
Have you got a chance to review base idea of this change?
It looks for me, that after changes connected with decimal (#871) it will be finished from my point of view.
This PR will be starting point for numeric type implementation of `IConvertible` interface (mostly works already in my local version) and using System.Convert class from mscorlib (except Base64 conversion, that needs  #860 before).

iskiselev:
Rebase on to of current tip. Removed Box for decimal.

iskiselev:
@kg, what else should be done here?

kg:
I should find time to review this until I feel okay merging it :-))))))))))))))

iskiselev:
Sorry for pushing, but I'd like to finish this, as I already have lot's of changes on top of it:
1. IConvertible interface implementation for raw types
2. Char moved to Number simple type - should be discussed, as have some issue
3. Nearly finished: Pointer operations with string
4. Nearly finished: interface methods for Arrays (removed ArrayOverlay)

It will be very sadly if I will need huge rework on all of that based on this PR discussion :(

Issue #871, opened 2015-10-14
Decimal values should be created through constructor


iskiselev:
JSIL use special type for handling decimal values (separate from Number), but still use raw Number if decimal created through `JSNumberLiteral`.
We should fix it. @kg which option do you prefer:
1. Replace JSNumberLiteral<decimal> with constructor call as step of pipeline (I suppose something similar happens with int64?)
2. Create JSDecimalLiteral and return it from JSLiteral.New(decimal) with special method in JavascriptAstEmitter.
3. Just write special logic inside JavascriptAstEmitter.VisitNode (JSNumberLiteral number).

kg:
JSDecimalLiteral sounds best.

PR #873, opened 2015-10-15
Create decimal literals through decimal constructor.

iskiselev:
Use constructor to create decimal literals (#871). When we implement full decimal support we should change this, as we still load value through conversion to double.

### DIFF
--- a/JSIL/AST/JSLiteralTypes.cs
+++ b/JSIL/AST/JSLiteralTypes.cs
@@ -244,6 +244,20 @@ namespace JSIL.Ast {
         }
     }
 
+    public class JSDecimalLiteral : JSLiteralBase<Decimal>
+    {
+        public JSDecimalLiteral(decimal value)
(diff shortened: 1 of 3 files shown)


JSIL/AST/JSLiteralTypes.cs
```diff
@@ -244,6 +244,20 @@ select em
         }
     }
 
+    public class JSDecmalLiteral : JSLiteralBase<Decimal>
```
kg: Typo

iskiselev:
Typo fixed. Removed Decimal case from GetActualType for JSNumberLiteral.

iskiselev:
Correct typo fix :)

iskiselev:
Please restart failed CI test. Is there anything else here to be discussed?

kg:
I didn't see anything wrong with it. Thanks.

Issue #846, opened 2015-09-11
Array type should implement interfaces.


iskiselev:
Each T[] type in .Net implements a number of interfaces:
```
System.ICloneable
System.Collections.IList
System.Collections.ICollection
System.Collections.IEnumerable
System.Collections.IStructuralComparable
System.Collections.IStructuralEquatable
```

One dimensional arrays should also implement:
```
System.Collections.Generic.IList<T>
System.Collections.Generic.ICollection<T>
System.Collections.Generic.IEnumerable<T>
System.Collections.Generic.IReadOnlyList<T>
System.Collections.Generic.IReadOnlyCollection<T>
```

Here is test case for it:
```
using System;
using System.Collections;
using System.Collections.Generic;
using JSIL;

public static class Program {
    public static void Main ()
    {
        var base1d = typeof (Base[]);
        TestType(base1d);
        TestTypeGeneric<Base>(base1d);
        TestTypeGeneric<object>(base1d);
        TestTypeGeneric<Derived>(base1d);
        TestTypeGeneric<Test>(base1d);

        var base2d = typeof(Base[,]);
        TestType(base2d);

        var struct1d = typeof(Struct[]);
        TestType(struct1d);
        TestTypeGeneric<Struct>(struct1d);
        TestTypeGeneric<object>(struct1d);
    }

    public static void TestType(Type t)
    {
        foreach (var @interface in t.GetInterfaces())
        {
            Console.WriteLine(@interface.Name);
        }

        Console.WriteLine(typeof(IEnumerable).IsAssignableFrom(t) ? "true" : "false");
        Console.WriteLine(typeof(ICollection).IsAssignableFrom(t) ? "true" : "false");
        Console.WriteLine(typeof(IList).IsAssignableFrom(t) ? "true" : "false");
        //Console.WriteLine(typeof(ICloneable).IsAssignableFrom(t) ? "true" : "false");
        //Console.WriteLine(typeof(IStructuralEquatable).IsAssignableFrom(twoD) ? "true" : "false");
        //Console.WriteLine(typeof(IStructuralComparable).IsAssignableFrom(oneD) ? "true" : "false");
    }

    public static void TestTypeGeneric<T>(Type t)
    {
        Console.WriteLine(typeof(IEnumerable<T>).IsAssignableFrom(t) ? "true" : "false");
        Console.WriteLine(typeof(ICollection<T>).IsAssignableFrom(t) ? "true" : "false");
[...]

Issue #844, opened 2015-09-09
Members of Object in reflection defined twice


iskiselev:
All members of System.Object in __ReflectionCache__ and type members are defined twice. Here is test case:
```
using System;

public static class Program {
    public static void Main ()
    {
        Console.WriteLine(typeof(object).GetConstructors().Length);
    }
}
```

Expected result : 1
Actual result: 2

iskiselev:
For object type it initially apply members from JSIL.ImplementExternals for "System.Object", and after it  members from JSIL.MakeClass for "System.Object" applied as new members.
Easy fix for it - remove `$.ExternalMethod` from `JSIL.MakeClass(Object, "System.Object"`, but probably some other type suffer from the same bug. I'm not sure that I'm ready to do proper fix - as type creation/initialization still is a magic for me.

PR #859, opened 2015-09-30
Upload build artifacts

iskiselev:
Here I changed travis scripts so that it use travis cache for Nuget/Compiled tests and is able to upload build artifacts (binaries and libraries, compiled tests and translated tests).

To use it:
1. Register [Yandex.Disk](https://disk.yandex.com/invite/?hash=OWD2151Q)
2. Get OAuth key: open [link](https://oauth.yandex.com/authorize?response_type=token&client_id=3dfa760cfe3d4e34bf9f6d8cd98b9de6) when logged in your Yandex.Disk.
4. Create secured environment variable on Travis-CI with name "JSIL_STORAGE_KEY"  and value from step 2.
5. Create "JSIL" folder in Yandex.Disk. Optionally share it (my folder shared r/o with everyone - https://yadi.sk/d/vd1C9YigjN5jt).
6. Yandex.Disk has Web.Dav api, so it can even be attached as network drive in windows (https://yandex.com/support/disk/webdav/webdav-win.xml). There will be 10 Gb of disk space.

PR #874, opened 2015-10-16
Misc reflection

### LINKED ISSUE #509: JSIL doesn't preserve name of method generic argument
JSIL currently doesn't preserve any generic information from generic method argument - it save only it's position number. So, if I have method `TOutput SomeMethod<TOutput>()`, I cannot get TOutput name from reading open generic return type name.

### LINKED ISSUE #847: Type for untyped array should be object[]
Currently JSIL for untyped array return System.Array, which is bad if reflection is used on it.
It should use object[] instead.

iskiselev:
Lot's of small changes and fixes (mostly reflection).
- Name for GenericParameter match .Net (both positional and type) - fix #509 
- System.Reflection.Runtime* types added.
- System.Reflection.RuntimeMethodInfo.GetParentDefinition implemented
- Array.Empty(), Array.UnsafeCreateInstance(...)
- Implemented hiding for Properties/Events/Constructors (reflection).
- SemaphoreSlim added to BclProxies as it was added to javascript implentation
- Assembly.Load implemented.
- RuntimeType.op_Equality implemented
- Thread.MemoryBarrier	
- MethodInfo.GetBaseDefinition and test added.
- MemberInfo.get_IsPrivate 
- Enum.ToObject fixed and additional signature added.
- Type.GetProperty additional override added (still need proper implementation)
- Use BCL implementation for System.Attribute with translated BCL (only for .Net, don't work with mono for now. Removed null-returning Attribute method from JS implementation.
- Enum.GetUnderlyingType added.
- Char.IsNumber added.
- JSIL.StringFromCharArray with passed length implementation fixed.
- MethodBase.IsConstructor and MethodInfo.IsVirtual added.
- Type.EmptyTypes added (with hack - it's property, not field.)
- Array marked as IEnumerable (part of #846).
- Array objects type is object[] now (fix #847, test added).
- Workaround for #844. Still should be investigated more. Test added.
- Sting.Length added (reflection compatibility).


iskiselev:
Don't understand why it have failed on AppVeyor. Probably threading issue on JSIL.mscorlib translation? It passed in my repo.
@kg, please run AppVeyor tests again.

kg:
Yeah, it looks like the codegen for the stubbed mscorlib was wrong somehow? However, it's an erroneous reference to an undeclared var, which is weird for stubbed code...

iskiselev:
BTW, if you'll review/merge #859 and register yandex.disk we'll have artifacts after each build for investigation (at least for travis, but I may add same script for AppVeyor also).

JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Array.js
```diff
@@ -172,4 +172,11 @@ JSIL.ImplementExternals("System.Array", function ($) {
       reverseImpl(array, index, length);
     }
   );
+
+  $.Method({ Static: true, Public: true }, "Empty",
+    new JSIL.MethodSignature($jsilcore.TypeRef("System.Array", ["!!0"]), [], ["T"]),
+    function Empty(T) {
+      return [];
```
kg: This needs to be a typed array for various primitive types.

JSIL.Libraries/Includes/Core/Reflection/Classes/System.Type.js
```diff
@@ -180,7 +183,11 @@
       new JSIL.MethodSignature(memberArray, [$jsilcore.TypeRef("System.Reflection.BindingFlags")]),
       function (flags) {
         return JSIL.GetMembersInternal(
-          this, flags
+          this,
+          flags,
+          undefined,
```
kg: I generally don't like seeing undefined literals, because they're indistinguishable from an uninitialized variable being passed to a function call. Is undefined the only correct thing in these cases?

JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Array.js
iskiselev: fixed.

JSIL.Libraries/Includes/Core/Reflection/Classes/System.Type.js
iskiselev: Really here i've added additional parameter to function. Here I wanted to set this added parameter, but still skip parameters that was undefined before (something like `method(arg1) - > method(arg1, undefined, newArg3)`), but in this case it is safe to path null. fixed, let's wait for CI results.

callanh:
Hi, I am new to JSIL and GIT so this could be a rookie question.  I've run into this problem:

The external method 'System.Object ToObject(System.Type, System.Object)' of type 'System.Enum' has not been implemented.

Can you advise if this is meant to be implemented and available?  One of the points in the topmost comment above is 'Enum.ToObject fixed and additional signature added.'

iskiselev:
@callanh, you have three options:
- Wait until that PR will be merged (I hope not too long)
- Checkout any version after 4551dda60a37a1e6f3b1bdc8da79f165ba379183 from [my fork](https://github.com/iskiselev/JSIL/tree/Misc_Reflection)
- Apply change from 4551dda60a37a1e6f3b1bdc8da79f165ba379183  to version that you have already manually (probably easiest way)

Issue #866, opened 2015-10-11
Usage of type through reflection before JSIL.InitializeType


iskiselev:
Looks like, if first type usage is through reflection (it could be easy when constructing types from open generic), we don't call JSIL.InitializeType.
I've noticed this as `typeOfValue.__AssignableTypes__` was null when `IsAssignableFrom` was called.

Will provide test case later.

kg:
Thanks, good point. We may just want to force initialization in those methods. The right way to do that would be to expose a public accessor method (GetAssignableTypes etc) much like with some of the other internal data, and the accessor ensures that the type is initialized.

iskiselev:
Test case:
```
using System;

public static class Program {
    public static void Main (string[] args) {
        Console.WriteLine(typeof(IInterface).IsAssignableFrom(typeof(Holder).GetField("Field").FieldType) ? "true" : "false");
    }
}

public interface IInterface {}
public class GenericClass<T> : IInterface
{ }

public class Holder
{
    public GenericClass<object> Field;
}
```

PR #879, opened 2015-10-21
Initialize type if not yet on IsAssignableFrom.

iskiselev:
Workaround for type initialization (#866). Test case added.

### DIFF
--- a/JSIL.Libraries/Sources/JSIL.Core.js
+++ b/JSIL.Libraries/Sources/JSIL.Core.js
@@ -5635,6 +5635,10 @@ JSIL.MakeType = function (typeArgs, initializer) {
       };
     } else {
       typeObject._IsAssignableFrom = function (typeOfValue) {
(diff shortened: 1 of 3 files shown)


JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -5635,6 +5635,10 @@ JSIL.MakeType = function (typeArgs, initializer) {
       };
     } else {
       typeObject._IsAssignableFrom = function (typeOfValue) {
+        if (!typeOfValue.__TypeInitialized__ && true) {
```
kg: Any reason why the true is here?

iskiselev: Probably too many work :) Really it is nearly same as ||false, but will
change to common form tomorrow.

kg: No need to change it, just wondering if anything special was going on there. Thanks.

Issue #880, opened 2015-10-21
Foreach closures and variable locality with anonymous delegates


callanh:
I may have found a significant bug!  An important pattern (to me at least) is declaring anonymous delegates inside a foreach.  The real use case is for dynamic UI generation.  Imagine the reproduction case below is foreach creating buttons with the click event handled.

```C#
using System;

static class Program
{
  static void Main(string[] args)
  {
    var List = new List<Action>();

    foreach (var Index in new int[] { 1, 2, 3 })
    {
      var Local = Index;

      List.Add(() => Console.WriteLine("Delegate Index(" + Index + ") vs Local(" + Local + ")"));
    }

    foreach (var Action in List)
      Action();
  }
}
```

In this example, it will print to the console:

```
Delegate Index(3) vs Local(3)
Delegate Index(3) vs Local(3)
Delegate Index(3) vs Local(3)
```

When it needs to print:

```
Delegate Index(1) vs Local(1)
Delegate Index(2) vs Local(2)
Delegate Index(3) vs Local(3)
```

Interestingly, JSIL is giving the same behaviour as once defined in the C# standard. You may remember that Microsoft retrospectively changed the implementation in the C# 5.0 Standard.

kg:
That's interesting. I wonder if this is ILSpy decompiling into a foreach and me implementing the old semantics.

kg:
When I run that snippet on ```try.jsil.org``` it produces the desired output. Any ideas? Maybe mono's C# compiler is producing something that decompiles correctly?

If you can send me a .dll or .exe that contains your compiled version of this test case, that would help me make sure I can reproduce it here. (I will be trying it out locally at some point)

callanh:
I've sent you via email because it won't let me post an .exe file.

iskiselev:
Looks like I've broken this through PR #807.

callanh:
Out of interest, the pre C# 5.0 Standard behaviour would actually have been:

```
Delegate Index(3) vs Local(1)
Delegate Index(3) vs Local(2)
Delegate Index(3) vs Local(3)
```

Note that the Local variable was a workaround for the foreach variable not being scoped in a useful way.

iskiselev:
There is nothing with C# standard. Problem is inside calling of delegate inside JSIL after my last fix. Hope to find solution and fix it soon.

iskiselev:
Issue could be closed.

Issue #876, opened 2015-10-18
Typo in JSIL.Core.js


callanh:
I believe that the plus '+' below should be a comma ','.  Fails with a format string exception instead of the warning message.

JSIL.Core.js:1165

```javascript
JSIL.WarningFormat(
            "Private type '{0}' with external implementation defined more than twice: '{1}'" + 
            [$private.toString(), existing.toString()]
```

kg:
You're right. Thanks.

iskiselev:
Sorry, please reopen this and close #875 instead. Looks like I've there was typo in commit message.

PR #893, opened 2015-10-29
Delagate fixes

iskiselev:
Here is new delegate invocation implementation.
Now I store all required info (method owner, name, signature, if call is virtual) in JSIL.MethodPointerInfo, and use appropriate call through signature in JS part.

It fixes both #876 and #880.

Future work options (we may discuss it here, but will post them in separate PR):
1. Create methods in runtime to exclude `.apply` usage in delegate invocation (at least don't use it twice).
2. Probably move delegates created in run-time from MethodInfo to the same approach. MethodInfo invocation probably also worth to change same way to unify codebase.

Note: Looks like Mono was updated on Travis, and two tests fails now. It is not related to this PR.

### DIFF
--- a/JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Delegate.js
+++ b/JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Delegate.js
@@ -109,35 +109,7 @@ JSIL.ImplementExternals("System.Delegate", function ($) {
       if (typeof (delegatePublicInterface.New) !== "function")
         JSIL.Host.abort(new Error("Invalid delegate type"));
 
-      if (!method.IsStatic && firstArgument == null && delegateType.GetMethod("Invoke").GetParameters().length == method.GetParameters().length + 1) {
-        //TODO: Check that first delegate Invoke parameter is assignable to method.DeclaringType
-        return delegatePublicInterface.New(
-          null,
-          function () {
-            var realFirstAgument = arguments[0];
-            var impl = JSIL.$GetMethodImplementation(method, firstArgument);
-
-            if (!impl) {
-              throw new System.Exception("Method has no implementation");
-            } else if (typeof (impl) === "function") {
-              var f = delegatePublicInterface.New(realFirstAgument, impl, function () { return null; });
-              return f.apply(null, Array.prototype.slice.call(arguments, 1));
-            } else {
-              throw new Error("Unexpected JSIL.$GetMethodImplementation result");
-            }
-          },
-          function () { return method; });
-      }
-
-      var impl = JSIL.$GetMethodImplementation(method, firstArgument);
-
-      if (!impl) {
-        throw new System.Exception("Method has no implementation");
-      } else if (typeof (impl) === "function") {
-        return delegatePublicInterface.New(firstArgument, impl, function () { return method; });
-      } else {
-        throw new Error("Unexpected JSIL.$GetMethodImplementation result");
-      }
+      return delegatePublicInterface.New(firstArgument, null, new JSIL.MethodInfoMethodPointerInfo(method));
     }
   );
 
@@ -169,14 +141,12 @@ JSIL.ImplementExternals("System.Delegate", function ($) {
       if (this.__isMulticast__) {
         return this.get_Method();
       }
-      if (!this.__isMethodInfoResolved__) {
-        var methodInfo = this.__methodInfoResolver__();
+      if (this.__methodPointerInfo__) {
         // TODO: find better solution for resolving MethodInfo in class by MethodInfo in base class.
         // Currently it will not find proper derived implementation MethodInfo for virtual method and interface methods.
-        JSIL.SetValueProperty(this, "__methodInfo__", methodInfo);
-        this.__isMethodInfoResolved__ = true;
+        return  this.__methodPointerInfo__.resolveMethodInfo();
       }
-      return this.__methodInfo__;
+      return null;
     }
   );
 });
--- a/JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Delegate.js
+++ b/JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Delegate.js
@@ -109,7 +109,7 @@ JSIL.ImplementExternals("System.Delegate", function ($) {
       if (typeof (delegatePublicInterface.New) !== "function")
         JSIL.Host.abort(new Error("Invalid delegate type"));
 
-      return delegatePublicInterface.New(firstArgument, null, new JSIL.MethodInfoMethodPointerInfo(method));
+      return delegatePublicInterface.New(firstArgument, null, JSIL.MethodPointerInfo.FromMethodInfo(method));
     }
   );
 
--- a/JSIL.Libraries/Sources/JSIL.Core.js
+++ b/JSIL.Libraries/Sources/JSIL.Core.js
@@ -10785,139 +10785,65 @@ JSIL.MethodPointerInfo.prototype.resolveMethodInfo = function () {
 }
 
 JSIL.MethodPointerInfo.prototype.createInvocationFunction = function(thisObject) {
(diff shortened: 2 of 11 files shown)


iskiselev:
Added function generation for delegate invocation.

JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Delegate.js
```diff
@@ -109,35 +109,7 @@ JSIL.ImplementExternals("System.Delegate", function ($) {
...
-      var impl = JSIL.$GetMethodImplementation(method, firstArgument);
-
-      if (!impl) {
-        throw new System.Exception("Method has no implementation");
-      } else if (typeof (impl) === "function") {
-        return delegatePublicInterface.New(firstArgument, impl, function () { return method; });
-      } else {
-        throw new Error("Unexpected JSIL.$GetMethodImplementation result");
-      }
+      return delegatePublicInterface.New(firstArgument, null, new JSIL.MethodInfoMethodPointerInfo(method));
```
kg: Wow, MethodInfoMethodPointerInfo sure is a name.

A method pointer constructed from a MethodInfo, I assume?

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -10732,4 +10739,207 @@ JSIL.Array.IndexOf = function (array, startIndex, count, value) {
...
+
+JSIL.MethodPointerInfo.$createStaticInvocation = function (argsCount, saveThis)
+{
+  var key = "InvokeStatic" + (saveThis ? "Attached" : "") + argsCount;
+
+  if (!(key in JSIL.MethodPointerInfo)) {
+    var innerArgs = [];
+    var innerBody = [];
+
+    innerBody.push("//# sourceURL=jsil://closure/JSIL.MethodPointerInfo." + key + "\r\n");
```
kg: Is there a good reason why this isn't using CreateNamedFunction?

JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Delegate.js
iskiselev: Yes, you are right:)

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -10732,4 +10739,207 @@ JSIL.Array.IndexOf = function (array, startIndex, count, value) {
...
+  this.IsVirtual = isVirtual;
+  this.MethodInfoResolved = false;
+  this.MethodInfo = null;
+
+  var useGenericSuffix = this.MethodGenericParameters !== null && this.MethodGenericParameters.length > 0;
+
+  this.NameWithGenericSuffix = useGenericSuffix ? (name + "$b" + this.MethodGenericParameters.length) : name;
+}
+
+JSIL.MethodInfoMethodPointerInfo = function(methodInfo) {
```
kg: Since this type has no methods of its own, please change this to a static method of JSIL.MethodPointerInfo, like ```JSIL.MethodPointerInfo.FromMethodInfo(mi)```.

JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Delegate.js
iskiselev: If we construct it from MethodInfo, we may cache method info at construction time to not do costly resolution operation later.

JSIL.Libraries/Sources/JSIL.Core.js
iskiselev: It creates function with closure arg stored inside on each call. I want create closure function once and reuse it.

JSIL/AST/JSExpressionTypes.cs
```diff
@@ -3256,23 +3285,6 @@ public JSNullCoalesceExpression (JSExpression lhs, JSExpression rhs, TypeReferen
         }
     }
 
-    public class JSDeferredExpression : JSExpression {
```
kg: Yay :D

JSIL.Libraries/Sources/JSIL.Core.js
iskiselev: It's brilliant idea! Thanks. Not sure why I haven't found such simple solution.

JSIL.Libraries/Sources/JSIL.Core.js
kg: I'm not sure I understand. ```CreateNamedFunction``` creates a function (optionally with closure values bound in) and returns it. You can also use [```CreateRebindableNamedFunction```](https://github.com/sq/JSIL/blob/eebc933ed7d2dd9cc5ecf1539d6ff9b5f3995edc/JSIL.Libraries/Sources/JSIL.Core.js#L3314), which returns the function you *call* to create the closure and pass it a set of values.

In both cases if you don't pass closure values it just creates a named function.

iskiselev: I need here outer function that will store closure variable. So let's see this:
```
function outer(closureArg) {
  return function inner(arg) { return closureArg + arg;};
}
```
If I call `CreateRebindableNamedFunction` it will still return me `inner` function. I need get `outer` function here, so that I'm able to reuse it.

iskiselev: Maybe I just don't understand how  `CreateRebindableNamedFunction` works. Will play a little bit more with it, but I when I tried today, I was unable to achieve desired behavior.

kg: Rebindable returns the outer function. I think it should be able to do what you want.

JSIL.Libraries/Sources/JSIL.Core.js
iskiselev: Fixed JSIL.MethodPointerInfo.FromMethodInfo.

kg:
I think I'm OK with merging this, but I'd like to see if you can get it using RebindableNamedFunction. Let me know if you can't.

JSIL.Libraries/Sources/JSIL.Core.js
iskiselev: Looks like CreateRebindableNamedFunction works. Thank you, will change it.

iskiselev:
I'd suggest change API for `JSIL.CreateRebindableNamedFunction`, so that `closure` will be array. It will be convenient to work.

JSIL.Libraries/Sources/JSIL.Core.js
iskiselev: Done. Use JSIL.CreateRebindableNamedFunction now (last arg is closure arg names array now).

iskiselev:
Simplified/unified methods that were used for invocation function generation.

iskiselev:
@kg, could you restart travis (223.2 and 223.3) - in first group there is predictable of two failed tests due to mono upgrade.

iskiselev:
@kg, should we move MethodInfo.Invoke to same MethodPointerInfo engine?

kg:
That sounds like a good idea, so Invoke is more consistent with everything else.

iskiselev:
OK, will do it when have some free time and prepare separate PR.

iskiselev:
Is there anything that still need to be reviewed?

kg:
I don't think so. What's up with the mono tests? It's just a bug in mono?

iskiselev:
I've not investigated it good enough, but build that was successful before is failing with same tests when I restart it now.

iskiselev:
Could check and probably mark them as FailsOnMono in separate PR.

iskiselev:
Checked once again: https://travis-ci.org/iskiselev/JSIL/jobs/86542137
The same build was passed successful 16 days before. Will mark this test as FailsOnMono.

kg:
OK. I'll merge.

PR #905, opened 2015-11-14
Raw types implements IConvertible

iskiselev:
Implementation of IConvertible interface for numeric types.
Interface method signature of IConvertible now able call methods on raw types.

Fixed bug in JSIL.InitializeType. Before, when derived type was initialized during base type initialization, it may skip some members from base type.


iskiselev:
Probably it worth skip this and go straight to #906. On other side both of #905 and #906 has huge changes and it may be easily to review them separately.

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -4651,7 +4707,7 @@ JSIL.RunStaticConstructors = function (classObject, typeObject) {
   // Run any queued initializers for the type
   var ti = typeObject.__Initializers__ || $jsilcore.ArrayNull;
   while (ti.length > 0) {
-    var initializer = ti.unshift();
```
kg: oh wow

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -8344,7 +8410,11 @@ JSIL.InterfaceMethod = function (typeObject, methodName, signature, parameterInf
   this.parameterInfo = parameterInfo;
   this.qualifiedName = JSIL.$GetSignaturePrefixForType(typeObject) + this.methodName;
   this.variantInvocationCandidateCache = JSIL.CreateDictionaryObject(null);
-  this.fallbackMethod = JSIL.$PickFallbackMethodForInterfaceMethod(typeObject, methodName, signature);
+  if (interfaceMemberFallbackMethod !== null) {
+    this.fallbackMethod = interfaceMemberFallbackMethod;
```
kg: piping it through like this is a good solution, i think.

Issue #904, opened 2015-11-13
Question: JSIL.InitializeType, __PreInitMembrane__, RunStaticConstructors


iskiselev:
@kg, I need some consultation from you.
I still not understand, how __PreInitMembrane__ works and when it is created. I'm also not sure, why RunStaticConstructors for some type called from InitializeType and only before first instance created for others.
Why we can't rely on membranes for types, that initialized from JSIL.Initialize?

I'm trying to solve some issues with cycle in type constructions, but feels that I need some more info on questions that I've asked here.

kg:
PreInitMembrane exists so that a type's static constructor runs when the type's members are first used. Otherwise, the static constructor would run when the type's public interface was first accessed, which is far too early and would cause cycles/stack overflows. It's basically a hack to approximate how static constructors work in .NET.

Membranes could probably be applied more universally. There are also some better solutions to type construction cycles possible but I've never been clear-minded enough to implement one without breaking everything.

iskiselev:
In what cases type static ctor could be executed right from InitializeType (through call to `classObject.__PreInitMembrane__.maybeInit();`)?

kg:
Before the introduction of the membrane, the only option was to run the cctor while creating the public interface, because after we created the public interface all the methods were accessible.

The membrane ensures that the methods aren't invoked until the cctor has run.

The correct fix for this is for ```asm.Namespace.Type``` to not expose the public interface, or something. That way all the cyclic dependencies and initialization order problems go away. That makes interop tough, though...

kg:
Also, I misread your comment, and I have no idea how maybeInit could invoke the cctor. It's probably a bug, maybe it's invoking the membrane handler when creating a membrane...

iskiselev:
Next question what is the purpose of next line inside `JSIL.InitializeType`:
```
      if (classObject.__PreInitMembrane__)
        classObject.__PreInitMembrane__.maybeInit();
      if (classObject.prototype && classObject.prototype.__PreInitMembrane__)
        classObject.prototype.__PreInitMembrane__.maybeInit();
```

This lines execute static ctors (for some classes).

kg:
I really don't remember what that does. If I get some time I'll read through things again.

iskiselev:
Thanks, will try to debug it a little bit more and to guess this myself :)

iskiselev:
Calling type static constructor from `JSIL.InitializeType ` removed through PR #906.

Issue #553, opened 2014-07-10
Variance with value types


iskiselev:
Per MSDN (http://msdn.microsoft.com/en-us/library/dd799517.aspx):
"Variance applies only to reference types; if you specify a value type for a variant type parameter, that type parameter is invariant for the resulting constructed type."

It doesn't restricted by JSIL libraries. Here is test case:

```
using System;

public static class Program
{
    public static void Main(string[] args)
    {
        TestType(new GenericClass<object>());
    }

    public static void TestType(object input)
    {
        if (input is IGeneric<A>)
        {
            Console.WriteLine("Variance!");
        }

        Console.WriteLine();
    }
}


public interface IGeneric<in T> { }
public class GenericClass<T> : IGeneric<T> { }

public struct A { }
```

PR #906, opened 2015-11-14
Array types reimplemented

### LINKED ISSUE #865: Interface implementations for raw types
This issue relates to #846, #787, #241, #864.
We should decide, how we will provide interface method implementations for raw JS types objects.
If I'm correct, here is full list of such objects:
1. Numbers: Byte, SByte, Int16, UInt16, Int32, UInt32
2. Boolean
3. Char
4. String
5. Array

Here is full list of interface that they implement:
1. (All except array) IComparable, IConvertible, IComparable<T>, IEquatable<T>
2. (Numbers) IFormattable
3. (String and arrays) ICloneable, IEnumerable, IEnumerable<char>
4. (Arrays) IList, ICollection, IStructuralComparable, IStructuralEquatable, IList<T>, ICollection<T>, IReadOnlyList<T>, IReadOnlyCollection<T>

I see several approaches for it:
1. Use JSIL.Box<T> for calling interface methods as proposed in #864. This solve problem for structs, but we still have problem with arrays and strings. We may introduce something similiar for them (as already done with JSIL.ArrayInterfaceOverlay), but it introduce reference equals problem - see #787.
2. Apply JSReplace attribute to all members of interfaces, that will use some static helper with passing this as first argument. May need additional tweaks for calls through delegate/reflection (see #850 as example).
3. Attach required methods to Array/String prototype - looks like a dirty hack for me.

@kg, could you suggest some additional options? What is you opinion? It looks for me that (2)
 option is best.

### LINKED ISSUE #895: 2d-array GetEnumerator() not implemented
Object of type 'JSIL.MultidimensionalArray' has no implementation of I83$GetEnumerator

```C#
  static class WebProgram
  {
    static void Main(string[] args)
    {      
      var SquareArray = new int[2, 2] { { 1, 2 }, { 3, 4 } };
      var SquareEnumerator = SquareArray.GetEnumerator();
      while (SquareEnumerator.MoveNext())
        Console.WriteLine(SquareEnumerator.Current);
    }
  }
```

### LINKED ISSUE #903: Generic method resolution for compound generic broken
Test case:
```
using System;
using JSIL.Meta;

public static class Program {
    public static void Main (string[] args) {
        var s = "a";
        OverloadedMethod(ref s);
        OverloadedMethod(new[] { "a", "b" });
        OverloadedMethod(new G<string>());
    }

    [JSRuntimeDispatch]
    public static void OverloadedMethod<T> (ref T t) {
        Console.WriteLine("OverloadedMethod<{0}>(ref <T>)", typeof(T));
    }

    [JSRuntimeDispatch]
    public static void OverloadedMethod<T> (T[] t) {
        Console.WriteLine("OverloadedMethod<{0}>(<T[]>)", typeof(T));
    }


    [JSRuntimeDispatch]
    public static void OverloadedMethod<T>(G<T> t)
    {
        Console.WriteLine("OverloadedMethod<{0}>(<G<T>>)", typeof(T));
    }
}

public class G<T>
{
}
```

Last call `OverloadedMethod(new G<string>());` will fail. Root cause here:
```
JSIL.ResolvedMethodSignature.prototype.ResolvePositionalGenericParameter = function (genericParameterValues, parameter) {
  if (
    (typeof (parameter) === "object") &&
    (parameter !== null) &&
    (Object.getPrototypeOf(parameter) === JSIL.PositionalGenericParameter.prototype)
  ) {
    return genericParameterValues[parameter.index] || null;
  } else {
    return parameter;
  }
};
```

`ResolvePositionalGenericParameter ` resolves only JSIL.PositionalGenericParameter, but really should try to resolve any open type.

iskiselev:
This PR is really logical continuation of #905. Some changes from #905 was improved here.

Array type JS implementation rewritten, so it looks more as .Net (we now have separate types for Array and Vector), array doesn't need ArrayOverlay anymore.  Array type variance correctly implemented.

Struct types now are non-variant, as in .Net. (#553, test __not__ added).

Type static ctor execution from intialliaztion removed (#904).

Fixed #903. Probably fixed #895 (to check separately).

Partially resolved #865 and related issues (need to gather tests from them).


iskiselev:
Looks like I've done something incorrectly while worked on #903. Will fix it.

iskiselev:
Renamed "JSIL.Core" assembly to "mscorlib".

iskiselev:
Issue fixed, all tests are passing now.

JSIL.Libraries/Includes/Core/Types/Classes/System.Array.js
```diff
@@ -1,129 +1,400 @@
...
-    $.RawMethod(false, "GetLength", function () {
-        return this.length;
-    });
-    $.RawMethod(false, "GetLowerBound", function () {
-        return 0;
-    });
-    $.RawMethod(false, "GetUpperBound", function () {
-        return this.length - 1;
+  $.RawMethod(false, "GetLength", function (dimension) {
+    if (this.LowerBounds || false) {
```
kg: ????????????

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -4546,81 +4581,119 @@ JSIL.InitializeType = function (type) {
...
+      }
+      baseType = baseType.__BaseType__;
+    }
+
+    JSIL.IsInsideInitializeType = true;
+    typeObject.__TypeInitializing__ = true;
 
-  if (typeObject.__IsClosed__) {
-    var forceLazyMethodGroups = false;
+    if (typeObject.__IsClosed__) {
```
kg: Do you know what happened here? Looks like a whitespace change?

JSIL.Libraries/Includes/Core/Types/Classes/System.Array.js
iskiselev: if object has this.LowerBounds, it means that we work with multi-dimensional array and should use info from specified dimension.

Proxies/Array.cs
```diff
@@ -21,25 +21,43 @@ public abstract class ArrayProxy {
         [JSAlwaysAccessAsProperty]
         [JSNeverReplace]
         [JSIsPure]
-        abstract public long LongLength { get; } 
+        abstract public long LongLength { get; }
 
 
+        [JSReplacement("$$jsilcore.System.Array.prototype.GetLowerBound.call($this, $dimenstion)")]
+        public int GetLowerBound(int dimenstion)
```
kg: ```dimenstion``` -> ```dimension```

JSIL/Transforms/ReplaceMethodCalls.cs
```diff
@@ -314,7 +314,7 @@ public class ReplaceMethodCalls : JSAstVisitor {
                             return;
                         }
                     }
-                } else if (
+                } /*else if (
```
kg: mind killing the commented-out code instead? Seems like it isn't needed anymore.

Proxies/Array.cs
```diff
@@ -21,25 +21,43 @@ public abstract class ArrayProxy {
...
+            throw new InvalidOperationException();
+        }
+
         [JSReplacement("JSIL.Array.New($elementType, $size)")]
         public static System.Array CreateInstance (Type elementType, Int32 size) {
             throw new InvalidOperationException();
         }
 
-        [JSReplacement("JSIL.MultidimensionalArray.New.apply(null, [$elementType].concat($sizes))")]
+        [JSReplacement("JSIL.MultidimensionalArray.CreateInstance(elementType, $sizes)")]
```
kg: I'm confused how one of these is MultidimensionalArray.CreateInstance and the following ones are .New.

kg:
This looks pretty good. Can you give me a quick summary of what the intended result of all the array changes is? Some of the detailed looked subtle so I want to make sure I understand the big picture.

JSIL.Libraries/Sources/JSIL.Core.js
iskiselev: try { was added to line 4589, so indentation was increased. Yes, diff in whitespace.

JSIL.Libraries/Includes/Core/Types/Classes/System.Array.js
kg: I see. Could you introduce a helper method (called ```IsMultidimensional``` or something) and call that instead? That will make it clearer what's being tested. It looks like a copy-paste error as-is.

Proxies/Array.cs
iskiselev: Current there is two helper functions for JSIL.MultidimensionalArray creation.
-New accept dimension length as [lowerBound1, length1, lowerBound2, length2, ...)
-CreateInstance accept [length1, length2] and use 0 for all lowerBound.

iskiselev:
Main purpose was to get rid of ArrayOverlay, but still support correct interfaces.
Original multi-dimension array implementation was partial - now it much more complete.
Arrays (multidimensional arrays) and vectors (one-dimension zero based array) have different implemented interfaces in .Net, so I need better separation between them.
Also, for each .Net array dimension there is separate type - now we have same in translated source.
Also I've worked to support array variance (MyClass[] is not implement IList<object>, but object still be cast to it as it could be cast to object[]).

Main principal change - all arrays created through .Net has attached property __ElementType__ (and we treat arrays without them as object[]).

Proxies/Array.cs
kg: OK. Please come up with a less subtle naming distinction. One of those two should probably make it explicit what they take so the other one's arguments are implicit by contrast.

kg:
OK, that all sounds good. Thanks for doing the hard engineering work to make that happen. :D

iskiselev:
You also already noted, that interface now can pass fallback method and additional check type method.

Proxies/Array.cs
iskiselev: Really, I tried to mimic .Net method signatures.
Array type ctor (New):
A constructor that takes twice as many int32 arguments as there are dimensions of the
array. These arguments occur in pairs—one pair per dimension—with the first argument of
each pair specifying the lower bound for that dimension, and the second argument 

Array.CreateInstance:
Creates a multidimensional Array of the specified Type and dimension lengths, with zero-based indexing. The dimension lengths are specified in an array of 32-bit integers.

Could you suggest better names?

kg: Ah, I see. If you're mirroring stuff in the BCL then this is fine. Thanks.

JSIL.Libraries/Includes/Core/Types/Classes/System.Array.js
iskiselev: IsNotVector?

kg: 'Vector' doesn't mean anything to me in this context. Is there a particular definition of it you're using that is referenced somewhere in a comment?

iskiselev: Really I can use just JSIL.IsArray, as for multidimansional array we still create wrapper object (it worked the same before my PR).

iskiselev: Personally, I don't like very much name Multidimensional array, as it can be really single-dimension (nut not zero-based, or even zero based, if somebody really wants it).
In ECMA-335 they named them as Array and Vector.
Unfortunately, in C# (and in JSIL java part) we really name use name Array instead Vector and Multidimensional array instead of Array.

iskiselev:
Removed commented code and fixed typo, use !JSIL.IsArray check instead of .LowerBounds.

kg:
OK, so it sounds like a 'Vector' is a single-dimension array with indexes starting at 0? And everything else is an 'Array' in that terminology? Is that right?

Anyhow, ```!JSIL.IsArray``` seems like a sensible solution. Thanks.

kg:
Follow-up note: What do you think we should do about JSIL.IsArray? Having a method with that name return false for certain arrays that aren't Array-enough is pretty goofy. IsJSArray, maybe?

iskiselev:
Really it is a little bit tricky.
Per ECMA-335, "an array type shall be defined by specifying the element type of the array, the rank (number of dimensions) of the array, and the upper and lower bounds of each dimension of the array. ...
Exact array types are created automatically by the VES when they are required. ...
The VES creates one array type for each
distinguishable array type. In general, array types are only distinguished by the type of their
elements and their rank. However, the VES treats single dimensional, zero-based arrays (also
known as vectors) specially."
So, you can really specify different array types in IL (and it will be preserved during compilation), but at run time will be created vector if "it is single-dimension array with indexes starting at 0" and array otherwise.

iskiselev:
It even more tricky that I thought.
Single dimension array is assignable from vector (but vector is not derived from single-dimension array type). Probably once I will take one more round to do it properly.

iskiselev:
So probably JSIL.IsArray should return true even for non-vector array (and they should also be just raw JS arrays, we just attach bounds/dimensions info to them) and we need to introduce JSIL.Array.IsVector method to detect non-vector arrays.

PR #907, opened 2015-11-16
Fixed output with DCE enabled

iskiselev:
When DCE stripped interface, I missed to strip it from another interface implemented interfaces list. Here is fix and test case added.


iskiselev:
Please run CI build again?

PR #914, opened 2015-11-30
Implement String.Join(string, IEnumerable<string>)

mwh:
This String.Join overload was not declared, although JSIL.JoinEnumerable
would already do the job if asked. This commit connects the two and
updates the existing test to cover the IEnumerable<String> case as well.
The other String.Join overloads are untouched.

### DIFF
--- a/Proxies/String.cs
+++ b/Proxies/String.cs
@@ -99,6 +99,12 @@ namespace JSIL.Proxies {
             throw new NotImplementedException();
         }
 
+        [JSIsPure]
+        [JSReplacement("JSIL.JoinEnumerable($separator, $values)")]
+        public static string Join (string separator, IEnumerable<string> values) {
+            throw new NotImplementedException();
+        }
+
         [JSChangeName("length")]
         [JSAlwaysAccessAsProperty]
         [JSNeverReplace]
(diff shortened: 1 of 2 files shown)


kg:
Looks good, thanks! I'll merge once tests pass.

iskiselev:
Next tests are broken:
DelegateMethod_Issue604
DelegateRemove 
DelegateRemoveAdvancedWithLambda_Issue541 
StringAsIConvertible

@mwh, could you check why (probably we have some other problem, not directly related to thus commit). I may also take a look on it later.

mwh:
All of those tests succeed for me locally (although I haven't been able to get the harness actually running the tests, so that's just building+running manually).

The DelegateRemove tests error in the Travis report with:

      Expected: "ABC\nBC\nAC\nAB\nC\nA\nB\nAB"
      But was:  "ABC\nBC\nAC\nAB\nC\nA\nABC\nAB"
but the output I get here from both mono (4.0.1) and JSIL matches the "was", not the "expected". I don't know enough about the test suite to tell what that means.

iskiselev:
Intersting. Mono changed implementation of delegate substract operation. Probably it is even right, but for sure different from .Net:
```
        Action a = A;
        Action b = B;
        Action c = C;
        Action s = a + b + c + Console.WriteLine;
        s();                  //ABC 
        (s - a)();            //BC 
        (s - b)();            //AC 
        (s - c)();            //AB 
        (s - (a + b))();      //C 
        (s - (b + c))();      //A 
        (s - (a + c))();      //.Net:ABC; Mono: B

        s = a + b + a;
        (s - a)();            // AB
```

Will look for other issues.

kg:
Thanks

PR #917, opened 2015-12-03
Correct argument order for StringBuilder.Insert

mwh:
The internal implementation function expected (self, string, startIndex,
count), but the bound methods both passed (self, startIndex, string,
count). The implementation would then insert the index at (usually) the
start of the string, which was not the desired behaviour.


kg:
Thanks.

PR #919, opened 2015-12-03
Define Enumerable.{Zip,Skip,Last}

mwh:
Each of these methods is implemented and a test case added. Enumerable.Last has a special optimised case for lists, as the native implementation does, and the others produce a JSIL.AbstractEnumerable.

### DIFF
--- a/JSIL.Libraries/Includes/Bootstrap/Linq/Classes/System.Linq.Enumerable.js
+++ b/JSIL.Libraries/Includes/Bootstrap/Linq/Classes/System.Linq.Enumerable.js
@@ -169,6 +169,89 @@
         }
       );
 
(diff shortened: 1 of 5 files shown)


kg:
This looks good, but I'll want to do a more exhaustive read of the code first. Thanks!

mwh:
The O(1) special case in Last() in particular was more trouble than it was worth and would benefit from someone with a better understanding of the internals taking a look - I'm sure there's a better way of handling the different possibilities for .Count.

kg:
To be honest I might just do that the lazy JS way, and do a dynamic check for ```.get_Count``` and ```.Count``` and then do another one for ```.get_Item``` and ```[i]```

kg:
Thanks

Issue #920, opened 2015-12-03
Some tests fail on Travis-CI


iskiselev:
After last Mono update some tests fail on Travis-CI (see #914). We need to review them and probably mark with FailsOnMono.

Also looks like most mono fails happen on domain unload. I want to split stubbed/translated tests into different configurations and try to disable creating new domain for each test, as it may improve CI staibility.

kg:
Sounds good. Shame about the domain thing. Would it make sense to investigate running multiple nunit processes instead of using appdomains?

iskiselev:
Bug report for Mono:
https://bugzilla.xamarin.com/show_bug.cgi?id=13813 - AppDomains
https://bugzilla.xamarin.com/show_bug.cgi?id=36640 - Delegate.Method
https://bugzilla.xamarin.com/show_bug.cgi?id=36646 - Delegate.Remove

kg:
Thanks for filing those reports.

PR #924, opened 2015-12-05
Mono tests

iskiselev:
Per discussion in #713 set default value for UseThreads to false. I still preserved "nt" command line option - for now it won't do anything. I've not added any opposite options, if user really wants enable theading he can do it in .jsilconfig.

Added FailsOnMono to tests that don't work on Mono anymore (#920), removed from some which work now.

Enabled back quite mode for zipping build artifacts - as TravisCI not always collapses command output.


kg:
Thanks for doing this.

PR #926, opened 2015-12-05
Travis fixes

iskiselev:
1. Disabled using of AppDomains in tests. Looks like it much more stable without them.
2. Changed artifacts upload script - it is possible to configure upload folder name now, use build date suffix for uploaded artifacts to preserve artifacts of rerun.

With this change, I can configure uploading build artifacts to my Yandex.Drive (publicly accessible https://yadi.sk/d/YnPDF_zfkzAbH). @kg, are you OK for me to update SQ/JSIL Travis-CI configuration after this PR will be merged?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ before_cache:
   - mv "/tmp/JSIL Tests" /tmp/JSIL_Tests  
   
 after_script:
+  - export BUILD_FINISH_TIME=$(date --utc +%Y%m%d_%H%M%SZ)
   - true && [ ! "$JSIL_STORAGE_FOLDER" ] && export JSIL_STORAGE_FOLDER=JSIL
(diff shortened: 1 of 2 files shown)


kg:
Yeah, sounds good. Thanks for knocking this out.

iskiselev:
Some magic happens.
After removing of app domains I don't se any sgen faults anymore, but I nearly every build see incorrect encoding of JS console outputl. Our test treats JS output, that is really UTF8 as UTF16LE. Really, first symbol there is UTF16LE BOM. Probably it would be easy to detect in our test and decode it back. I've seen similar errors before several time, but not as often as it is happen without app domain usage :(

iskiselev:
Added fix to correct encoding. Looks like it works stable enough

iskiselev:
Updated Travis-CI settings, build artifacts are now accessible via https://yadi.sk/d/YnPDF_zfkzAbH

Really, it worse create some notes in wiki about different type of JSIL test-cases, how to run them, what CI we have and how can they be configured for fork.

PR #927, opened 2015-12-06
Implement variants on string operations, and File.OpenText

### LINKED ISSUE #646: string.Split( throws exception if count parameter is passed.
The following stops with a script error.  If you remove the second parameter on split it works fine.  I tests it with "A:B:C" as well to make sure it wasn't a one off problem.

            var sep = new char[] {':'};

            string str = "A:B";
            var split = str.Split(sep,2);
            Console.WriteLine(split.Length);

mwh:
These commits implement four methods:

* String.Split with a count. This partially fixes #646; if a count is provided, and there is only one delimiter, it will retrofit the correct semantics onto JavaScript split. If there is a count and multiple delimiters, it raises an error, in the same way that providing a StringSplitOptions argument does. The test case gains an extra check for split-with-count.
* String.StartsWith with a comparison type. This is *not* always correct, and I'm not sure if it's possible to be on the platform. It case-folds for all IgnoreCase comparison types, and uses the same indexOf check that the basic String.StartsWith does for Ordinal and other cases.
* Char.ConvertToUtf32(string, int). This just returns s.codePointAt(i) if it's available, and checks for surrogate pairs manually otherwise.
* File.OpenText. This just returns a new System.IO.StreamReader on the path.


JSIL.Libraries/Includes/Bootstrap/Text/Classes/System.Char.js
```diff
@@ -79,6 +87,36 @@ JSIL.ImplementExternals("System.Char", function ($) {
...
 
+  $.Method({ Static: true, Public: true }, "ConvertToUtf32",
+    new JSIL.MethodSignature($.Int32, [$.String, $.Int32], []),
+    function ConvertToUtf32(s, i) {
+      // If codePointAt is available, use it, but it isn't
+      // widespread enough to rely on yet.
+      if (s.codePointAt)
+        return s.codePointAt(i);
+      // Otherwise, do shifting arithmetic for surrogate pairs.
+      var hs = s.charCodeAt(i);
```
kg: Sorry to take so long to get back to you on this, but there's an implementation of this surrogate pair workaround in the text encoding code already, named ```$jsilcore.charCodeAt``` and ```$jsilcore.fromCharCode```.

JSIL.Libraries/Includes/Bootstrap/Text/Classes/System.Char.js
```diff
@@ -79,6 +87,36 @@ JSIL.ImplementExternals("System.Char", function ($) {
...
+        return (hs - 0xd800) * 0x400 + (ls - 0xdc00) + 0x10000;
+      }
+      return hs;
+    }
+  );
+
+  $.Method({ Static: true, Public: true }, "ConvertFromUtf32",
+    new JSIL.MethodSignature($.String, [$.Int32], []),
+    function ConvertFromUtf32(i) {
+      if (i >= 0x10000) {
```
kg: See above about implementations in ```$jsilcore```

mwh:
Thanks, I've updated the methods to use those functions.

Issue #932, opened 2015-12-10
Typed Array Cast/Clone


griestopf:
Whiile updating [FUSEE] (https://github.com/FUSEEProjectTeam/FUSE) to the latest JSIL version I encountered two issues with Typed Arrays. I could not reproduce them using tests in JSIL, on the other hand I cannot figure out what could be wrong from our side, so I describe the two issues here and also illustrate how I managed to workaround by changing code in JSIL.Core.js.

Typed Array Cast
----------------------
In our C# code we have the following statement:
```C#
      paramExisting = new EffectParam()
      {
           ShaderInxs = new List<int>(new int[] {i}),
      };
```
where `i` is an integer 

This JSIL-compiles to
```JavaScript
    var effectParam2 = new ($T10())();
      effectParam2.ShaderInxs = $S09().Construct($T1D().$Cast(JSIL.Array.New($T04(), [i])));
      effectParam.set(effectParam2);
      (this._parameters).Add(current.Name, effectParam.get());
```

Then the browser crashes in the `$Cast` which is in JSIL.Core.js (~5095)
```JavaScript
  castFunction = function Cast(expression) {
    if (isFunction(expression))
      return expression;
    else if (expression === null)
      return null;
    else
      throwCastError(expression);
  };
```
Here `expression` is a typed array. So it's neither a function nor is it null -> throwCastError is called.

I changed the JSIL.Core.js code to:

```JavaScript
  castFunction = function Cast(expression) {
    if (isFunction(expression) || JSIL.IsTypedArray(expression))
      return expression;
    else if (expression === null)
      return null;
    else
      throwCastError(expression);
  };
```
Which works well. I tried to write the following test, but inside JSIL's nunit test environment the test succeeds:

```C#
using System;
using System.Collections.Generic;
using JSIL.Meta;

public static class Program {
    public static void Main (string[] args) {

        Test[] test = new Test[5];

        for (int i = 0; i < 5; i++)
        {
            test[i] = new Test
                {
                    IntList = new List<int>(new int[] {i})
                };
        }

        for (int i = 4; i >= 0; i--)
            Console.WriteLine("i: " + test[i].IntList[0]);
    }
}

public class Test
{
    public List<int> IntList;
}
```

Typed Array Clone
------------------------
I noticed a crash in JSIL.Core.js (~9350) in the function `Array_New`:
```JavaScript
JSIL.Array.New = function Array_New (elementType, sizeOrInitializer) {
[...]

PR #933, opened 2015-12-11
Fixed Array.Clone with native JS arrays

iskiselev:
Fixed issue #932 and test added.


kg:
Always nice when a fix deletes code instead of adding it :-) Thanks

PR #935, opened 2015-12-15
JS-interop functions return object

iskiselev:
It is next part for #925 (and include that PR).
JS-interop return object instead of dynamic now.

Probably we need update samples used for TryJSIL, but I don't know where they are located.


kg:
What did you update the submodules to point to?

iskiselev:
With latest commit from sq/JSIL.Meta NoDynamic branch (b185a3d1a30e6d580456e6d78eaf07f0b06a13bb)

kg:
As long as it's not pointing towards JSIL.Meta master, can you also merge this into a branch on sq/JSIL? I want to switch all the master branches over wholesale at once.

iskiselev:
Sure. Will push it to branch NoDynamic.

iskiselev:
Done.

PR #957, opened 2016-03-25
Preparation for Cecil update

iskiselev:
Latest ILSpy reference Mono.Cecil as submodule. It will be really hard to make changes in Cecil (as it will be second level submodule) after we upgrade it.
Fortunately, we have only one change in Cecil: in BaseAssemblyResolver class. Really, we need this change only when we use JSIL.Internal.AssemblyResolver, so we could move it there. After it, we'll be able to use no-patched  Mono.Cecil version.

This PR is not fully equivalent to original change - I've added lock to all Resolve method instead of checking assembly path, but it should not affect us great.

### DIFF
--- a/JSIL/CILSupport.cs
+++ b/JSIL/CILSupport.cs
@@ -82,6 +82,9 @@ namespace JSIL.Internal {
             25, 52, 224, 137
         };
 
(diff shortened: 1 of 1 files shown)


kg:
Looks fine

PR #959, opened 2016-03-26
Improved source map generation

iskiselev:
With this PR source map generated by JSIL is much more usable.

### DIFF
--- a/JSIL/AST/JSAstVisitor.cs
+++ b/JSIL/AST/JSAstVisitor.cs
@@ -41,6 +41,7 @@ namespace JSIL.Ast {
         protected bool VisitNestedFunctions = false;
 
         protected event Action<JSNode> AfterNodeProcessed;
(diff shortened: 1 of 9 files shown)


iskiselev:
As with this PR source map generated by JSIL will be usable for most tasks, we probably should create separate wiki page with information about it or mention it somewhere else - to make more people know about this flag.

iskiselev:
Probably we may think how to integrate them into TryJSIL, so that end-user will have an option with which he'll be able to see/debug original C# source in browser developer tools.

kg:
Does this depend on any preceding PRs?

iskiselev:
No, This PR contains all required commits.

kg:
I assume this passes without the spidermonkey bug in effect?

iskiselev:
Yes, the only failed test is Spidermonkey's array sort.

PR #971, opened 2016-04-21
Web workers support

iskiselev:
Here is basic implementation of WebWorker environment support based on #656.
Also there JSIL.js was changed a little bit for better compatibility with JS packers (tested with browserify). When it pack files, each file is placed inside function which is called on newly created object. After it, resolving of global scope through this doesn't work anymore. However, in both browser and webworker global scope could be accessed with `self` property. So, if self is defined, it is used as global scope reference.


iskiselev:
I'll send updated version in few days.

Issue #772, opened 2015-07-30
Warning on Roslyn compiled struct field access


iskiselev:
While looking on #768, I've noted a lot of warnings: "Accessing {0} without a reference as this." ({0} - field full name).
Looks like next code:
```
public struct TestStruct
{
    private int _a;

    public bool IsEqual(TestStruct value)
    {
        return _a == value._a;
    }
}
```
compiled on Roslyn produce next IL:
```
        IL_0000: nop
        IL_0001: ldarg.0
        IL_0002: ldfld int32 TestStruct::_a
        IL_0007: ldarg.1
        IL_0008: ldfld int32 TestStruct::_a
        IL_000d: ceq
        IL_000f: stloc.0
        IL_0010: br.s IL_0012
        IL_0012: ldloc.0
        IL_0013: ret
```

Same, compiled on previous version:
```
	IL_0000: nop
	IL_0001: ldarg.0
	IL_0002: ldfld int32 TestStruct::_a
	IL_0007: ldarga.s 'value'
	IL_0009: ldfld int32 TestStruct::_a
	IL_000e: ceq
	IL_0010: stloc.0
	IL_0011: br.s IL_0013

	IL_0013: ldloc.0
	IL_0014: ret
```

So, it use `ldarg, ldfld`  instead of `ldarga, ldfld`  to access field of structure argument. At the same time, `ILBlockTranslator.IsInvalidThisExpression` treat value type arg as invalid this expression, so we see warning.

kg:
Good point. For some reason I didn't realize ldarg ldfld would be valid, but of course it is.

PR #976, opened 2016-04-28
Correct fix for warning on roslyn-compiled struct field access.

iskiselev:
Looks like yesterday in #973 I missed something. Here is corrected fix for #772.


kg:
Yeah, this makes sense.

Issue #970, opened 2016-04-18
Function local method signatures optimization


iskiselev:
All signatures that refers class/function generic parameter ends in function local method signatures (`var $s__ = new JSIL.MethodSignature(...);`).

This signatures often used for only one method call (and they are not cached), but they still use high-cost `JSIL.CreateNamedFunction`.
We should either put them in some signature-cache (same as all other function signatures) or make some special processing inside that will just use `call`/`apply` method inside in that case.

iskiselev:
Looks like `"CodeGenerator": { "CacheGenericMethodSignatures": false  }`, should disable them - but looks like it currently just crash on compiling with it.

kg:
Putting them in a shared cache with everything else sounds like a good idea.

iskiselev:
@kg, why do you use type references in non function local signatures instead of type identifiers?

kg:
I honestly don't remember. Sounds like it might just be an artifact of old code doing the wrong thing. Could be code reuse, yeah.

PR #977, opened 2016-04-29
Method signatures cache improvements

iskiselev:
Here i implemented caching of open generic signatures (#970) and preserving only one signature cache per signature (896).

Both features are hidden under default settings.
(1) To enable global caching for open generic signatures user should change config:
CodeGenerator.PreferLocalCacheForGenericMethodSignatures=false
(2) To enable global caching for open generic interface method signatures user should change config:
CodeGenerator.PreferLocalCacheForGenericInterfaceMethodSignatures=false
(3) To enable storing only one method signature cache per signature:
CodeGenerator.CacheOneMethodSignaturePerMethod=false

I will still work more on method calls performance improvements, as all suggested option could either increase or decrease execution time dependent on usage scenario.

For my project I get about 20% performance improvement with `CodeGenerator.PreferLocalCacheForGenericMethodSignatures=false`. All other settings are actually have not changed measured performance of my application.

### DIFF
--- a/JSIL.Libraries/Sources/JSIL.Core.js
+++ b/JSIL.Libraries/Sources/JSIL.Core.js
@@ -571,6 +571,26 @@ JSIL.Memoize = function Memoize (value) {
   };
 };
 
+JSIL.MemoizeTypes = function MemoizeTypes(storage, valueProvider, args) {
+  var argTypes = "";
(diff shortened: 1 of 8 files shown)


kg:
Cool, thanks for doing this. Sounds like a big improvement.

PR #984, opened 2016-05-11
System.Convert.ToBase64CharArray

iskiselev:
A little bit dirty `System.Convert.ToBase64CharArray` implementation based on existing `ToBase64String`

### DIFF
--- a/JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Convert.js
+++ b/JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Convert.js
@@ -83,6 +83,15 @@ JSIL.ImplementExternals("System.Convert", function ($) {
     return result;
   };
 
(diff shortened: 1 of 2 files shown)


JSIL.Libraries/Includes/Bootstrap/Core/Classes/System.Convert.js
```diff
@@ -83,6 +83,15 @@ JSIL.ImplementExternals("System.Convert", function ($) {
     return result;
   };
 
+  function toBase64CharArray(inArray, offsetIn, length, outArray, offsetOut, options) {
+      var tempArray = JSIL.StringToCharArray(toBase64StringImpl(inArray, offsetIn, length, options));
```
kg: Can you change toBase64StringImpl to take an output array instead so you don't need a temporary string like this?

PR #988, opened 2016-05-26
TypeScript DefinitelyTyped support

iskiselev:
Initial implementation of TypeScript DefinitelyTyped mappings for translated code.
It is implemented as separate emitter. To enable, jsilconfig should contain: `"EmitterFactories" : ["JavascriptEmitterGroupFactory", "DefinitelyTypedEmitterGroupFactory"]`

When enabled, for each assembly additionally created 3 files:
- `module.%assemblyName%.d.ts`
- `module.%assemblyName%.js`
- `internals\%assemblyName%.d.ts`

And additionally:
- `module.JSIL.Core.d.ts`
- `module.JSIL.Core.js`
- `internals\JSIL.Core.d.ts`

JS file is just a wrapper on top of `JSIL.GetAssembly(...)` in node.js module format.
To use type definition user should reference `module.%assemblyName%`.

What's to do next:
- Introduce test framework and cover with test cases
- Discuss if we need wrap IEmitterGroupFactory[] and IAssemblyEmmitterFactory[] into special types.
- Discuss way to remove values on merge from configuration list/dictionaries
- Better coverage for JSIL.Core.d.ts
- Better interfaces support
- Support generic constraints
- Support of generic interface covariance. Option to enable bivariance for contrvariant generic interfaces.
- Proxied members support.

### DIFF
--- a/JSIL/AssemblyTranslator.cs
+++ b/JSIL/AssemblyTranslator.cs
@@ -420,14 +420,15 @@ namespace JSIL {
         }
 
         protected virtual string FormatOutputFilename (string fileName) {
+            var name = Path.GetFileNameWithoutExtension(fileName);
             foreach (var filenameReplaceRegex in Configuration.FilenameReplaceRegexes) {
-                fileName = Regex.Replace(fileName, filenameReplaceRegex.Key, filenameReplaceRegex.Value);
+                name = Regex.Replace(name, filenameReplaceRegex.Key, filenameReplaceRegex.Value);
             }
 
             if (Configuration.FilenameEscapeRegex != null)
-                return Regex.Replace(fileName, Configuration.FilenameEscapeRegex, "_");
-            else
-                return fileName;
+                name = Regex.Replace(name, Configuration.FilenameEscapeRegex, "_");
+
+            return Path.Combine(Path.GetDirectoryName(fileName), name + Path.GetExtension(fileName));
         }
 
         public TranslationResultCollection Translate (
--- a/JSIL/AssemblyTranslator.cs
+++ b/JSIL/AssemblyTranslator.cs
@@ -101,10 +101,8 @@ namespace JSIL {
         public event Action<TypeIdentifier> ProxyNotMatched;
         public event Action<QualifiedMemberIdentifier> ProxyMemberNotMatched;
 
-        public Func<bool, string, string> ChooseCustomAssemblyName;
-
         public readonly TypeInfoProvider TypeInfoProvider;
-        public readonly IEmitterFactory EmitterFactory;
+        public readonly IEmitterGroupFactory[] EmitterGroupFactories;
 
         protected bool OwnsAssemblyDataResolver;
         protected bool OwnsTypeInfoProvider;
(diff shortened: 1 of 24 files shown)


electricessence:
Whoa... Ok... does this work now?  How can I see this in action?

iskiselev:
@electricessence, this PR is not yet merged, so to play with it you need to update JSIL from my from:
git checkout -b iskiselev-d.ts master
git pull https://github.com/iskiselev/JSIL.git d.ts

When you'll build JSIL with this change, you  need either add `"EmitterFactories" : ["JavascriptEmitterGroupFactory", "DefinitelyTypedEmitterGroupFactory"]` to *.jsilconfig that you use to translate or call JSIL with command-line parameters `-e JavascriptEmitterGroupFactory -e DefinitelyTypedEmitterGroupFactory`

iskiselev:
@kg, can you look through it and suggest what should be improved?
For now I haven't wrap array of IEmitterGroupFactory and IAssemblyEmmitterFactory into some special class - they are really passed very few time.

JSIL/AssemblyTranslator.cs
```diff
@@ -514,47 +528,47 @@ struct MethodToAnalyze {
...
+                        if (!Manifest.GetExistingSize(assembly, emmitterFactory.Id, out existingSize)) {
+                            using (var outputStream = new MemoryStream(DefaultStreamCapacity)) {
+                                var sourceMapBuilder = Configuration.BuildSourceMap.GetValueOrDefault() ? new SourceMapBuilder() : null;
+                                var context = MakeDecompilerContext(assembly.MainModule);
+
+                                try {
+                                    var tw = new StreamWriter(outputStream, Encoding.ASCII);
+                                    var formatter = new JavascriptFormatter(tw, sourceMapBuilder, this.TypeInfoProvider, Manifest, assembly, Configuration, stubbed);
+                                    var emitter = emmitterFactory.MakeAssemblyEmitter(formatter);
+                                    TranslateSingleAssemblyInternal(emitter, context, assembly, outputStream, sourceMapBuilder);
```
kg: Oh, this is super weird. Why are we translating the whole assembly once per emitter? Is there a reason why it has to work that way? It seems like it might be much better to do a single translation that calls out to all the emitters.

Otherwise, there's a risk that each individual translation could produce a slightly different result and the emitter outputs wouldn't match up.

iskiselev: I believe that real transformation happens in this function before - when it calls `AnalyzeFunctions` and `RunTransformsOnAllFunctions`.
Despite it's name TranslateSingleAssemblyInternal only writes output. Probably I'm not correct.

kg: Hmm, you are right that most of the work happens before. Maybe a formal split (with a rename) is sufficient to make it work out okay. I'll make some time to review the code and see whether the split is correct or not.

iskiselev:
To make overload interface methods usable from typescript typed code, we need implement RuntimeDispatch overload selection for interface methods.

iskiselev:
@kg, now to call non-generic interface method user should run:
`Interface.IntrefaceMethod.Call(thisArg, null, arg1, arg2, ...)`
and for generic:
`Interface.IntrefaceMethod.Call(thisArg, [TArg1, TArg2,...], arg1, arg2, ...)`

At the same time calling class methods (with RuntimeDispatch) is much easier:
`thisArg.Method(arg1, arg2, ...)`
and for generic:
`thisArg.Method(TArg1, TArg2,...)(arg1, arg2, ...)`

I suggest make interface method (with newly-introduce RuntimeDispatch)  callable with next syntax:
`Interface.IntrefaceMethod(thisArg)(arg1, arg2, ...)`
and for generic:
`Interface.IntrefaceMethod(thisArg)(TArg1, TArg2,...)(, arg1, arg2, ...)`.

Later we may introduce special function: `Interface.$UnWrap(thisArg)`, that will return such object that user call any interface members on it directly:
`Interface.$Wrap(thisArg).Method(arg1, arg2, ...)`
and for generic:
`Interface.$Wrap(thisArg).Method(TArg1, TArg2,...)(arg1, arg2, ...)`

We may also introduce method `Interface.$Wrap(thisArg)`, that will accept raw js object and return valid JSIL interface-derived object, with all methods delegate to js-object method - it will provide ability to easy implement some typed object passed into JSIL from TypeScript.

kg:
```Interface.$Wrap``` makes sense, though I wonder how we would make it maintain all the other important properties.

iskiselev:
I thought, that I will generate class, implementing .Net interface, that accept raw js object on constructor and delegate each method to appropriate method of raw object. `$Wrap` function will generate such type (if it was not yet generate) and create it's instance using passed-in object.

kg:
One common problem with wrapping/unwrapping approaches that shows up in APIs like boost.python is that you can end up with multiple unique wrappers surviving for a given object, which breaks referential equality. Breaking referential equality is pretty nasty.

iskiselev:
It will break reference equity, as it just create new object. We can do a little trick and use WeakMap to store created wrapped/unwrapped object, so you'll have guaranty, that if you have two objects, created by same interface wrap/unwrap method from same object they will be equal.

iskiselev:
Added interface method runtime dispatcher and typing for interface methods (JSIL.Core.js changed for it).

Issue #991, opened 2016-05-31
Proposal: QualifiedMethods


iskiselev:
Based on my work on #988 and thoughts on #185, #623 I'd like to propose next:
Let's introduce QualifiedMethod. Qaulified method will 
- store type reference and method name. It may also contain method signature
- extend function so on call will invoke target method either through signature or runtime dispatcher:
  - for instance method will accept `this` reference and return function, that
  - for generic methods will accept generic arguments and  return function, that
  - will accept method arguments and call target method
- have `Call` method, that will accept this reference, generic argumetns and method arguments in single call for perfomance-optimised calls
- have `Of` method, that will accept same arguments, as MethodSignature and return QualifiedMethod with stored MethodSignature.
- be stored in type public interface for static methods (same as now stored static method invocators, but they will be fully compatible) and inside $InstanceMethod property of type public interface for instance method (here they will replace InterfaceMethods that currently stored in type public interface).

Such organisation will allow us later solve #185, #623, #990 and provide statically-typed way to call any method with specified signature from TypeScript code.

kg:
This sounds fine design-wise, but I'd be concerned about the performance implications. Any thoughts on how we can minimize how many layers of function calls it adds and cut down on any indirection?

iskiselev:
Such method in initial implementation may add only one additional layer to normall call through signature (and will not add anything if compare with call through InterfaceMethod). Direct call of such objects will be nearlt same as current method call without signature.

Really, it may even give us some benefits - if we change our method signature caching algorithms to always use them with in-function caching, our function invocation cache will be always (except variance interface calls) switch with one case and created after first method call.

iskiselev:
I'll add do some benchmarks to understand better if it affects call perfomance.

iskiselev:
I've implemented extended InterfaceMethod to work with all non-static functions (static will add later). Each type now holds InterfaceMethods for each instance method. Using them to call methods add no overhead comparing to calling method through signature - and even have some benefits, as each InterfaceMethod correspond with only one method overload. So, we don't need use inline cache anymore for all except variance calls.
With it I was able to solve #185, #623 (so InheritedAndNewMethod and VirtualAndNewMethods tests are not broken anymore).
But it come with price: we should use signature call a little bit oftener and class initialization takes longer time (have not measured how much).

Issue #623, opened 2014-10-29
Incorrect method call for non-virtual methods.


iskiselev:
Here is test case:

```
using System;

public static class Program
{
    public static void Main()
    {
        Base d = new Derived();
        d.I1();
    }
}

public class Base
{
    public void I1()
    {
        Console.WriteLine("base");
    }
}

public class Derived : Base
{
    public new void I1()
    {
        Console.WriteLine("derived");
    }
}
```

Expected result: `base`. Actual result: `derived`.

Issue #624, opened 2014-10-31
Delegate with detached instance method


iskiselev:
In .Net, you can create delegate to instance method that will accept object on which you want call it as first delegate call argument. It is not supported in JSIL. Here is small test case:

```
using System;
using System.Runtime.InteropServices;

public static class Program
{
    public static void Main()
    {
        var d = (Action<ClassWithInstanceMember>)Delegate.CreateDelegate(typeof(Action<ClassWithInstanceMember>), null, typeof(ClassWithInstanceMember).GetMethod("Run"));
        d(new ClassWithInstanceMember(1));
        d(null);
    }
}

public class ClassWithInstanceMember
{
    public int _value;

    public ClassWithInstanceMember(int value)
    {
        _value = value;
    }

    public void Run()
    {
        if (this != null)
        {
            Console.WriteLine("this: " + _value);
        }
        else
        {
            Console.WriteLine("this==null ");
        }
    }
}
```

I will work on this issue, as it is requirement for Expression Interpreter work with instance methods. Full test case should check:
- Non-generic methods in non-generic class
- Generic methods in non-generic class
- Non-generic methods in generic class
- Generic methods in generic class
- Non-generic methods in non-generic interface
- Generic methods in non-generic interface
- Non-generic methods in generic interface
- Generic methods in generic interface

PR #993, opened 2016-06-09
Qualified methods

### LINKED ISSUE #185: Implement full IL .overrides and virtual semantics
The way .overrides and virtual work in IL, you can actually have multiple chains of virtual methods with the same name on an object, by using 'new' to establish a new chain. It's also possible to use .overrides to override ordinary virtual methods instead of just interface methods.

C# and VB.net don't use either of these features extensively (though it's possible to do the chains using a mixture of 'new' and 'virtual') but it is probably necessary to implement this at some point anyway.

iskiselev:
This PR is not yet finished, but I'd like to start discussions for it. Main goal here is implement #991 and fix #185, #623.
Instead of MethodSignature for method calls used InterfaceMethod (that is just signature+name+source type).

Also fixed huge number of small bugs that was revealed during implementation (mostly with very edge cases, so they was not noticed before).


iskiselev:
As @kg supposed, this PR really increase memory usage.
I've compared my app compiled with `CacheOneMethodSignaturePerMethod` before this PR and after it. Here is memory usage according to Chrome Take Heap Snapshot:
Before: 358Mb
After: 508Mb

I've not supposed that memory usage increase will be so big. Will be glad for any help/suggestions what we can do with it

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -3014,6 +3041,26 @@ JSIL.InstantiateProperties = function (publicInterface, typeObject) {
 $jsilcore.CanFixUpEnumInterfaces = false;
 
 JSIL.FixupInterfaces = function (publicInterface, typeObject) {
```
iskiselev: Main changes is inside this method. It now calculates proper virtual member implementation for each base type, not only implemented interfaces.
Check is stricter now - method should be virtual and for auto-binding by name to interface class must explicitly declare that interface in it's own declaration.

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -8367,50 +8522,170 @@ JSIL.ResolvedMethodSignature.prototype.toString = function () {
...
-  result.fallbackMethod = this.fallbackMethod;
+  if (this.__OfCache__ == null) {
+    JSIL.RuntimeError("Rebind of signature-aware InterfaceMethod");
+  }
+
+  var result = new JSIL.InterfaceMethod(newTypeObject, this.methodName, newSignature, this.fallbackMethod);
+  return result;
+};
+
+JSIL.InterfaceMethod.prototype.RegisterSignature = function (signature, skipExistingCheck) {
```
iskiselev: All possible signatures should be registered inside InterfaceMethod now.

JSIL.Libraries/Sources/JSIL.Core.js
```diff
@@ -8367,50 +8522,170 @@ JSIL.ResolvedMethodSignature.prototype.toString = function () {
...
+      signature.toString(this.methodName)]);
+  }
+
+  var signatureAwareInterfaceMethod = new JSIL.InterfaceMethod(this.typeObject, this.methodName, signature, this.fallbackMethod);
+  signatureAwareInterfaceMethod.__OfCache__ = null;
+
+  this.__OfCache__[cacheKey] = signatureAwareInterfaceMethod;
+};
+
+JSIL.InterfaceMethod.prototype.Of = function (signature) {
```
iskiselev: We can get InterfaceMethod with specific signature, that was previously registered inside. It will be used instead of calling Signature with passing InterfaceMethod as a name.

JSIL/AssemblyTranslator.cs
```diff
@@ -1316,18 +1316,15 @@ Cachers cachers
         }
 
         internal JSFunctionExpression TranslateMethodExpression (
```
iskiselev: Passing MethodReference everywhere was artifact of times, when we translated closures inside parent function. We don't need it anymore, so I removed them and changed MethodReference/TypeReference to MethodDefinition/TypeDefinition when possible. It simplifies reading and understanding greatly.

iskiselev:
@kg, why do we cache MethodSignatures/TypeRefernce per type instead of caching them per assembly?

kg:
@iskiselev Mostly because it was easier. I see no reason not to do it assembly-wide.

iskiselev:
I've broke #624 with null arg to instance method with this change. I believe it's OK, as nobody should really write such code. We still may fix it later.

iskiselev:
Looks like I'm nearly done with it. The only left is moving Call method creation from MethodSignature to InterfaceMethod and small cleanup of code inside MethodSignature / InterfaceMethod.
What should be one later (not inside this PR, but when I'll have more time to work on it):

- Cleanup of translator code. We need to review usage of CachedQualifiedSignatureRecord/CachedSignatureRecord. Really CachedSignatureRecord used only for constructor invocation now. Remove lot's of Cache settings flag that are not used anymore. Remove BaseMethodCacher. Probably rewrite CachedQualifiedSignatureRecord to replace node with special CachedNode.
- Review invocation of method from generic type. We really could use non-resolved form in signature for all, except interfaces.
- Cache generic methods locally in-function.

iskiselev:
@kg, I think we could not use MethodSignature/InterfaceMethod Call/CallNonVirtual/CallStatic at all.
Instead we'll use methodKey from InterfaceMethod. So, instead of:
```
QS01().Call(thisReference, [gArg1, gArg2], arg1)
QS01().CallNonVirtual(thisReference, [gArg1, gArg2], arg1)
QS02().CallStatic([gArg1, gArg2], arg1)
```
We will use for non-variance call:
```
thisReference[QS01().Key](gArg1, gArg2, arg1);
thisReference[QS01().NonVirtualKey](gArg1, gArg2, arg1);
(T01()[QS01.Key])(gArg1, gArg2, arg1) // here we may return back using just MethodSignature
```
And for variance calls:
```
thisReference[QS01().LookupVariantMethodKey(thisReference)](gArg1, gArg2, arg1);
```
I believe it should not have any performance issues, but will improve performance/memory footprint as we will not create Call method closures.

What do you think?

kg:
It looks gross, but it might be considerably better for performance, so maybe it's worth it. It does solve the this-reference binding problem and it eliminates most of the need for closures, which is nice.

iskiselev:
I hate JavaScript! You can get performance problem nearly anywhere.
I've tested my idea with using MethodKey.
Here is example: http://plnkr.co/edit/hVsmbE?p=preview
I've rewrite in plain JavaScript  our a little bit simplified BaseMethodCall performance test. There Accessor is something similar to my last idea, Signature - as it worked with Call using InterfaceMethod function, and Prototype - how non-virtual call works before. 
So, here is result:
.Net: 47ms
Chrome | Accessor: 580 ms first calls, 120 ms next calls (not always, sometimes it stays 580 forever)
Chrome | Signature: 75 ms
Chrome | Prototype : 20 ms
Firefox | Accessor: 210 ms
Firefox | Signature: 11 ms
Firefox | Prototype : 11ms
Edge | Accessor: 800-900 ms
Edge | Signature: 970 ms
Edge | Prototype: 70 ms first calls, 25 ms next calls

@kg, have you any ideas what we could do with it?

iskiselev:
Looks like the only way to achieve normal performance would be calculate function overload name using some stable algorithm and hardcode them on translation. Otherwise we will be much slower comparing to .Net.

iskiselev:
Results for stable:
Chrome | Stable: 580 ms first calls, 130 ms next calls (not always, sometimes it stays 580 forever)
Firefox | Stable: 22 ms
Edge | Stable: 650-1050 ms

iskiselev:
We definitely need report bug to Chackra, as both Signature and Stable takes too long, Even Prototype takes longer, then others. I just not sure how it should be described.

kg:
The solution here is probably (ugh) to move to shipping some sort of bytecode in Release builds, and then translate the bytecode to JS at loadtime. That translated JS can have all the keys hardcoded as ```x["y"]``` or ```x.y```. I already wanted to do this and have some plans (there's a design Issue open about it) but I don't know when I'll have the energy to actually work on it.

iskiselev:
@kg, tried create small sample with run-time generated functions and added it to sample:
Chrome | Stable: 22 ms
Firefox | Stable: 33 ms
Edge | Stable: 850-1050 ms

As I can see, Chakra works bad with run-time generated functions (will create simplified sample and report, if it is not reported yet). It's interesting for me, why Firefox works 3 times slower comparing to  Signatures case. But still it works fast enough to try it.
I think I'll be able to hack our AST Emitter, so it will create stings with run-time calculated inserts instead of function signatures. It's also interesting what should we do with type cache. Probably it is used not so often, or we may create global dictionary that will hold resolved type by string type it.

kg:
My long-term strategy for bytecode would be to move things around such that the existing ASTEmitter can run in the browser, and then we ship a serialized version of the JS AST to it over the wire. For debug builds we would just run the ASTEmitter up front and save the output .js files.

iskiselev:
It really sounds great, but I need to solve problem right now. Looks like hacking AST would be the fastest way to do it.

iskiselev:
Safari also likes Stable/Generated approaches. They run faster than any else.

kg:
If this is blocking you right now, it sounds like a short-term solution is
in order. I'm not really sure what makes the most sense here.

I'm okay with a stable compile-time name generation approach as long as we
can make it comprehensive. The reason I've been relying on doing it at
runtime is that there are a ton of edge cases to consider, and ILSpy/Cecil
(especially ILSpy) don't give us 100% trustworthy information to use for
generating names.

iskiselev:
Right now I'm planning use "Generated" approach. So, each function will be written as sting with includes of run-time calculated signatures. Something like:
```
    $.Method({Static:false, Public:true }, "Method", 
      JSIL.MethodSignature.Void, 
      new FunctionCreator(function() {return ""+
        "this[" + $asm00.Base.$Methods.Method.InterfaceMethod.Of(JSIL.MethodSignature.Void).methodKey + "]();" + "\r\n" +
        "JSIL.Types[" + $asm00.Program.__TypeId__ +"].I = (((JSIL.Types[" + $asm00.Program.__TypeId__ +"].I | 0) + 2) | 0);"})
    );
```

kg:
Yikes. Well, as long as it's an option... I'd like to find a better
solution than that.

iskiselev:
Really I'll use it only for release build, but if it gives me 2-5 times performance improvement I'll accept it. If performance improvement will be less than 2 times, I'll not use it.

kg:
Yeah if it's that much faster it's worth it. Ideally we can just improve
the quality/readability of that code a little. Like instead of big string
concats, a simple template/regex replacement operation done by
FunctionCreator. Then the function bodies can at least be single string
literals (or lists of strings, one per line).

As-is all that concatenation means you can't effectively ctrl-f in the
output sources and it's much harder to breakpoint them. I'm not sure how it
would interact with source maps, either.

iskiselev:
Templates would not help us with ctrl-f, as we plan to modify string.
I've thought about sourcemaps. We really can create sourcemap back to js file from generated function easy enough, if we additionally encode info about line where function was originally defined. It is also possible to make sourcemap back to .net sources, but it will take some more work.

iskiselev:
If we will talk about "Stable" names approach, the main problem now is that I can't found way to encode function for interfaces. If we'll find such way, it would be also pretty good option.

PR #994, opened 2016-06-15
Variance support for d.ts

iskiselev:
Added variance support for *.d.ts declarations. Fixed some issue.


iskiselev:
It was really funny to find a way to describe interface generic argument covarinance and contrvarince in TypeScript :)

PR #1006, opened 2016-07-17
Implemented read-only support for System.IO.Compression.ZipArchive

dmirmilshteyn:
Adds basic support for reading zip files through System.IO.Compression.ZipArchive using [JSZip](https://stuk.github.io/jszip/) 2.x. It's dual licensed under MIT or GPLv3, so should be fine to use. Currently setup to be optional, and will only be loaded if the zip config option is set.


kg:
Can you add a couple simple test cases for this?
Thanks

PR #1010, opened 2016-07-19
Type.IsInstanceOfType implemented

iskiselev:
Type.IsInstanceOfType implemented.


kg:
It's nice when these fixes are like 3 lines :-)

PR #1012, opened 2016-07-20
Implemented Regex.IsMatch(string, string)

wklingler:


kg:
I'm not sure I like this, isn't ecmascript regex mode different from the .net default? This would break app logic

On July 20, 2016 2:50:57 PM PDT, Wade Klingler <notifications@github.com> wrote:
>
>You can view, comment on, or merge this pull request online at:
>
>  https://github.com/sq/JSIL/pull/1012
>
>-- Commit Summary --
>
>  * Implemented Regex.IsMatch(string, string)
>
>-- File Changes --
>
>M
>JSIL.Libraries/Includes/Bootstrap/Text/Classes/System.Text.RegularExpressions.Regex.js
>(9)
>
>-- Patch Links --
>
>https://github.com/sq/JSIL/pull/1012.patch
>https://github.com/sq/JSIL/pull/1012.diff
>
>---
>You are receiving this because you are subscribed to this thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/sq/JSIL/pull/1012

- kg (mobile)

kg:
Hmm, that's probably close enough. I'll think about merging it. Thanks!

wklingler:
Is there something that makes you nervous that I could look into more?

kg:
I'm pretty sure there are some cases where .NET regex syntax would parse but produce the wrong results in javascript, which could cause very confusing failures.
It would be ideal to write a .NET regex parser that rejects incompatible regexes. This might be OK though

wklingler:
That makes sense.  I'll see if I can come up with something like that sometime.

PR #1013, opened 2016-07-22
MyGet package publishing

iskiselev:
Implemented publishing JSIL package to MyGet feed. Use [GitVersion](http://gitversion.readthedocs.io/en/latest/) for versioning, bumped version to 0.9.1 for now.
JSILc from nuget package now able to copy Libraries to designed folder (use settings JsLibrariesOutputDirectory in jsilconfig or outLibraries command line parameter).

Current version of package doesn't contains any XNA translation related assemblies. I have no plans to integrate them for now (but it could be done by somebody else).

### DIFF
--- a/Compiler/Analyzers/DeadCodeAnalyzer/JSIL.DeadCodeAnalyzer.csproj
+++ b/Compiler/Analyzers/DeadCodeAnalyzer/JSIL.DeadCodeAnalyzer.csproj
@@ -49,6 +49,9 @@
     <Reference Include="System.Xml" />
   </ItemGroup>
   <ItemGroup>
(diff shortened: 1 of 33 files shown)


iskiselev:
It would be good to create ReadMe, that will be opened after package installation and will suggest user to create *.jsilconfig with at least next content:
```
{
  "OutputDirectory": "JS\\Translated\\",
  "JsLibrariesOutputDirectory": "JS\\Libraries\\"
}
```

iskiselev:
We also need to make JSIL.Meta package, but that package should be build from JSIL.Meta repository.

PR #1021, opened 2016-10-14
Added implementation for Tuple with up to 3 items

wklingler:
As far as I could tell Tuple is not implemented (I get Not Implemented errors when I use it). This isn't really a complete pull request.  I added the implementation for Tuple but I'm not sure if I did it the way you guys would like me to do it.  Could you just look over this and let me know if I'm on the right track and if I am I'll finish the implementation for up to 8 or 9 items.  Thanks


kg:
This seems right to me, can you add one or two tests for it? They don't need to be complicated.

wklingler:
I have implemented the Tuple class which can now create any tuple of any size if the developer follows the Tuple class creation rules for c#.  I also added in the Tuple.Create method which means the Tuple class should now be completely usable if I didn't miss anything.

Issue #1035, opened 2017-02-24
Assembly.GetType is not working for generics.


tosandeepgarg:
Following code doesn't work.
```c#
typeof(object).Assembly.GetType(
                "System.Collections.Generic.Dictionary`2[[System.Int32, mscorlib],[System.Object, mscorlib]]");
```

Here is test case:
```c#
using System;

public static class Program
{
    public static void Main(string[] args)
    {
        var typeInt32 = typeof(object).Assembly.GetType("System.Int32");
        var typeDictionary =
            typeof (object).Assembly.GetType(
                "System.Collections.Generic.Dictionary`2[[System.Int32, mscorlib],[System.Object, mscorlib]]");
        Console.WriteLine(
            "{0} {1}", typeInt32.Name, typeDictionary.Name
            );

    }
}
```

PR #1036, opened 2017-02-24
Assembly.GetType fixed

tosandeepgarg:
Issue #1035 was fixed. Test was added.


kg:
Thanks, glad it was a simple fix

PR #1040, opened 2017-03-01
Corrected generic type FullName property 

### LINKED ISSUE #1039: Type.FullName is not returning correct name for generic dictionary
Type.FullName returns generic arguments without separator.

Here is test case:
```c#
using System;
using System.Collections.Generic;

public static class Program
{
    public static void Main(string[] args)
    {
        var dictionary = new Dictionary<A,B>();
        var dictionaryType = dictionary.GetType();
        Type type = Type.GetType(dictionaryType.FullName);
        if (type != null)
            Activator.CreateInstance(type);
        Console.WriteLine(
            "{0},{1}", dictionaryType.Name,dictionaryType.FullName
            );

    }
}

public class A { }
public class B { }
```

tosandeepgarg:
We have fixed #1039. Test case added


kg:
Thanks!

PR #1042, opened 2017-03-08
Async improvements

mdaveynis:
Implementation of Task.FromResult<TResult>(TResult result).
Added support for task cancellation.


kg:
Thanks.

PR #1053, opened 2018-02-28
Implement Enumerable.Repeat

mwh:
The external method Repeat<TResult>(TResult,int) was not implemented.
This commit introduces a test case and matching implementation returning
a JSIL AbstractEnumerable returning the element the appropriate number
of times.

### DIFF
--- a/JSIL.Libraries/Includes/Bootstrap/Linq/Classes/System.Linq.Enumerable.js
+++ b/JSIL.Libraries/Includes/Bootstrap/Linq/Classes/System.Linq.Enumerable.js
@@ -557,6 +557,32 @@
         }
       );
 
+      $.Method({ Static: true, Public: true }, "Repeat",
+        new JSIL.MethodSignature(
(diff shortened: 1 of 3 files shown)


JSIL.Libraries/Includes/Bootstrap/Linq/Classes/System.Linq.Enumerable.js
```diff
@@ -557,6 +557,35 @@
...
 
+      $.Method({ Static: true, Public: true }, "Repeat",
+        new JSIL.MethodSignature(
+          $jsilcore.TypeRef("System.Collections.Generic.IEnumerable`1", ["!!0"]),
+          ["!!0", "System.Int32"],
+          ["TResult"]),
+        function Repeat$b1(TResult, element, count) {
+            var i = 0;
+
+            var moveNext = System.Collections.IEnumerator.MoveNext;
```
kg: This is unnecessary

mwh: Good point, I've taken them both out.

